PR #243, opened 2023-12-28
opt.:try add high refresh rate support

dccif:
On some Android devices, high refresh rates are not available. 

I was able to enable high refresh rate globally on my OnePlus 7 Pro by adding this [flutter_displaymode](https://pub.dev/packages/flutter_displaymode) module.


lollipopkit (approved):
Thanks for your contribution 😀

PR #384, opened 2024-06-11
chore: setup fastlane for F-Droid submission

Integral-Tech:
https://gitlab.com/-/snippets/1895688


lollipopkit:
image

It seems that `short_description.txt` should located in same the dir as `full_description.txt`?

Issue #381, opened 2024-06-10
Publish on F-Droid


Integral-Tech:
It would be better if ServerBox is submitted to F-Droid.

lollipopkit:
That's a good idea. Could you please help me submit this app to F-Droid, or should I do it myself?

lollipopkit:
Thank you for your efforts. :)
I have read the doc and it states that a Fastlane config should be created inside this repo.
I would like to get more people to contribute, so could you open a PR for that?

Integral-Tech:
> Thank you for your efforts. :) I have read the doc and it states that a Fastlane config should be created inside this repo. I would like to get more people to contribute, so could you open a PR for that?

Yes, I have opened a [pull request](https://github.com/lollipopkit/flutter_server_box/pull/384)

Integral-Tech:
In the [RFP issue](https://gitlab.com/fdroid/rfp/-/issues/2780), the bot detected non-free dependencies (seems to be GMS-related), which should be removed:
- com.google.android.c2dm.permission.RECEIVE

There are also some dependencies / services that could be used to track users, which should be minimized: 
- ly.count.android.sdk.messaging.CountlyPushActivity
- ly.count.android.sdk.ReferrerReceiver
- https://firebase.google.com/support/privacy/init-options
- https://plus.google.com/

lollipopkit:
It's related to #379, i will solve it later.

licaon-kter:
trying to build https://github.com/lollipopkit/flutter_server_box/commit/64ce3638cb8942499800875876d3a6fbe657516d

these proprietary libs are still pulled in via https://github.com/Countly/countly-sdk-flutter-bridge/blob/24.4.1/android/build.gradle#L38 and https://github.com/Rexios80/watch_connectivity/blob/watch_connectivity-0.2.0/watch_connectivity/android/build.gradle#L52

lollipopkit:
> trying to build [64ce363](https://github.com/lollipopkit/flutter_server_box/commit/64ce3638cb8942499800875876d3a6fbe657516d)
> 
> these proprietary libs are still pulled in via https://github.com/Countly/countly-sdk-flutter-bridge/blob/24.4.1/android/build.gradle#L38 and https://github.com/Rexios80/watch_connectivity/blob/watch_connectivity-0.2.0/watch_connectivity/android/build.gradle#L52

Thanks for the tip.

licaon-kter:
[...]

PR #385, opened 2024-06-11
chore: add distributionSha256Sum in gradle

Integral-Tech:
F-Droid suggests adding distributionSha256Sum in gradle/wrapper/gradle-wrapper.properties.
https://gitlab.com/fdroid/rfp/-/issues/2780


lollipopkit:
#381

PR #604, opened 2024-10-07
[Translation] Russian fix

MasedMSD:
- Where words began with a small letter in English, they now begin with a capital letter 
- Fixed some issues with incorrect word usage
- Changed `CPU`, `ЦПУ` and etc to `ЦП`

There may also be more problems with some of the sentences. I can't be sure that I've fixed absolutely everything


lollipopkit (approved):

lollipopkit:
Thanks for your fix.

PR #642, opened 2024-11-23
Delay bio auth

dsvf:
I find it difficult when switching between Server Box and other apps that the fingerprint auth activates immediately.

I implemented a delay, if switching away from Server Box and back to it within 10 seconds, then there is no auth request. I also added a setting under App/Android Setting and App/iOS Settings to enable and disable this behavior.

The Strings in the settings widget are hardcoded english since I did not want to add a second pull request to lppcg/fl_lib
Also, the 10 seconds are hardcoded and the "10" value in the settings subtitle is not synchronized with the delay in home.dart, since I'm not sure where and how you like to store a global constant in your project.

You are welcome to consider including this PR into the app.

## Summary by Sourcery

Add an optional delay for biometric authentication when returning to the app shortly after backgrounding it.

New Features:
- Introduce a setting to delay biometric auth lock for up to 10 seconds when switching back to the app
- Add a new toggle for delayed bio auth lock in Android and iOS settings pages

Enhancements:
- Track the app’s paused timestamp and conditionally skip authentication on resume if within the configured delay


denis0001-dev (approved):

lollipopkit:
@sourcery-ai review

PR #731, opened 2025-03-25
fix dist in issue717

moli765:
Sorry about that. There are some problems in Flutter analysis. So just fix it...

## Summary by Sourcery

Bug Fixes:
- Fixes a crash when the dist value is null.


lollipopkit:
Fixed

Issue #992, opened 2026-01-06
On Windows, SFTP file download fails with PathNotFoundException and RangeError


GT-610:
**Describe the bug**

SFTP file download functionality fails on Windows due to multiple path handling issues:

1. **Invalid folder name**: Uses `spi.oldId` (format: `user@ip:port`) as folder name, which contains characters illegal on Windows (`@` and `:`)
2. **Path separator mismatch**: Mixes Unix-style forward slashes (`/`) with Windows backslashes (`\`), resulting in malformed paths like `C:\Users\...\root@127.0.0.1:22\//root\compose`
3. **RangeError**: Substring operations use hardcoded `/` separator while `localPath` uses platform-specific separators, causing `lastIndexOf` to return -1

This only happens on Windows, as other platforms use `/` as seperator, which won't cause any issue.

**To Reproduce**

1. Run the application on Windows
2. Navigate to the SFTP file browser for a remote server
3. Attempt to download a file from the remote server
4. Observe the error messages in the logs

<img width="1332" height="628" alt="Image" src="https://github.com/user-attachments/assets/f1af43ce-2375-4111-8a5a-a28661a9c1ad" />

This error also exists in File page, while opening a folder.

<img width="1332" height="628" alt="Image" src="https://github.com/user-attachments/assets/b88eecd6-2219-49d4-aea6-35f89a6f3251" />

**Error Messages**

```
[package:server_box/main.dart:73] [WARNING] App: Exception: sftp worker event: PathNotFoundException: Exists failed, path = 'C:\Users\...\AppData\Roaming\ServerBox\file\root@127.0.0.1:22\\\root\compose' (OS Error: 文件名、目录名或卷标语法不正确。, errno = 123)
```

```
[package:server_box/main.dart:73] [WARNING] App: Exception: sftp worker event: RangeError (end): Invalid value: Not in inclusive range 0..95: -1
```

**Desired Results**

- Files should download successfully on Windows
- Folder names should use valid characters for all platforms (e.g., use `spi.id` instead of `spi.oldId`)
- Path separators should be consistent and platform-appropriate
- No RangeError exceptions during file operations

**Actual Results**

- File download fails with PathNotFoundException
- Invalid folder names are created on Windows
- Path separators are mixed, creating malformed paths
- RangeError occurs due to incorrect separator usage in substring operations

**Root Cause Analysis**

The issues stem from inconsistent path handling across the codebase:

[...]

PR #993, opened 2026-01-06
fix(sftp): Fix SFTP file download failures on Windows

GT-610:
Fix multiple path handling issues that caused SFTP file download failures on Windows:

1. **Invalid folder names**: Replaced `spi.oldId` (format: `user@ip:port`) with `spi.id` (format: `ptPQHje00`) for folder naming, as the former contains characters illegal on Windows (`@` and `:`)
2. **Path separator mismatch**: Fixed inconsistent use of path separators by normalizing all path operations to use platform-specific separators (`Pfs.seperator`)
3. **RangeError**: Fixed substring operations that used hardcoded `/` separator while `localPath` used platform-specific separators
4. **Triple separators**: Fixed path joining logic that resulted in malformed paths with triple separators (`\\\`)

This bug has been reported in #992.

### Modified Files

- `lib/view/page/storage/sftp.dart`
  - Changed `spi.oldId` to `spi.id` for folder naming
  - Added logic to strip leading separator from `normalizedPath` before joining paths

- `lib/data/model/sftp/worker.dart`
  - Changed hardcoded `/` to `Pfs.seperator` in `lastIndexOf` operation

- `lib/data/model/sftp/req.dart`
  - Changed hardcoded `/` to `Pfs.seperator` in `split` operation

This change only affects internal path handling logic and should not impact the public API.

The application already generates server IDs using `ShortId.generate()` (e.g., `ptPQHje00`), which are valid folder names on all platforms. The `spi.oldId` (format: `user@ip:port`) is still available for display purposes but should not be used for file system operations on Windows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file path handling to use platform-appropriate path separators across file and directory operations, ensuring proper functionality across different operating systems.

### DIFF
--- a/lib/data/model/sftp/req.dart
+++ b/lib/data/model/sftp/req.dart
@@ -36,7 +36,7 @@ class SftpReqStatus {
   late SftpWorker worker;
   final Completer? completer;
 
-  String get fileName => req.localPath.split('/').last;
+  String get fileName => req.localPath.split(Pfs.seperator).last;
 
   // status of the download
   double? progress;
--- a/lib/data/model/sftp/worker.dart
+++ b/lib/data/model/sftp/worker.dart
@@ -70,7 +70,7 @@ Future<void> _download(SftpReq req, SendPort mainSendPort, SendErrorFunction sen
     mainSendPort.send(SftpWorkerStatus.sshConnectted);
 
     /// Create the directory if not exists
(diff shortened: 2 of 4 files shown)


GT-610:
`Pfs.seperator` will auto-detect seperator of the current platform. For example, on Windows it will be `\`; on macOS, GNU/Linux, Android and iOS it will be `/`.

PR #996, opened 2026-01-08
feat(local file page): Display server names for server folders

GT-610:
On the File page, server folders were displayed using only their generated IDs (e.g., `ptQHjje00`), which are not user-friendly and difficult to identify.

QQ_1767860939004

This PR adds a feature that displays server folders with their configured names, with the ID shown as a subtitle:

- **Title**: Server name (e.g., "My Server")
- **Subtitle**: Server ID (e.g., `pxk9kk100`) in grey text

QQ_1767861126624

Folders of deleted servers (first one in the pic), non-servers will display their names as usual.

This only applies to folders directly under the root directory, avoiding false positives for nested folders with matching names.

## Summary by CodeRabbit

* **New Features**
  * Server-backed folders now display their server name as the title with the folder name shown as a subtitle for reference
  * Enhanced storage item display logic to properly distinguish between server-backed and local items

### DIFF
--- a/lib/view/page/storage/local.dart
+++ b/lib/view/page/storage/local.dart
@@ -143,7 +143,7 @@ class _LocalFilePageState extends ConsumerState<LocalFilePage> with AutomaticKee
     final isServerFolder = isDir && file.parent.path == Paths.file;
     String? serverName;
     if (isServerFolder) {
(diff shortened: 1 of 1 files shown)


lollipopkit:
Thanks :)

Issue #929, opened 2025-10-07
"Follow system" color seed gives the wrong colors to some elements 


MathisP75:
Some elements, such as the bottom navigation bar and the information cards, are the same color as the background.

Without using system colors: 
![Screenshot_20251006-213933.png](https://github.com/user-attachments/assets/9785fbf2-66db-44dc-8993-021933af9cda)

Using system colors:
![Screenshot_20251006-213700.png](https://github.com/user-attachments/assets/49bca22d-0a34-4f70-80fc-e315e11da5f9)

GT-610:
Seems that it only happens on Android. Investigating.

GT-610:
Got the problem. Trying to solve this.

GT-610:
Fixed.

<img width="1080" height="2424" alt="Image" src="https://github.com/user-attachments/assets/b7654082-6108-4edd-a5ee-6901ba1bc049" />

<img width="1080" height="2424" alt="Image" src="https://github.com/user-attachments/assets/af65c512-583e-495b-a36e-9b4ab34993a5" />

There are some other bugs in color picking, and I'll solve them together in one PR.

GT-610:
PR Merged. Let's just wait for the next release :)

lollipopkit:
#998

PR #998, opened 2026-01-10
fix(color-picking): Fix color picking failure and card overflow

GT-610:
- Fix issue #929
- Color change will be applied instantly after picking
- Fix card overflow by 2 pixels

## Summary by CodeRabbit

* **New Features**
  * Conditional dynamic color theming that follows system dynamic colors when available, with a saved-color-seed fallback.

* **Improvements**
  * Primary color editor pre-fills current hex, updates live with picker, validates and saves immediately.
  * "Follow System" option shown only when dynamic color is available.
  * Slightly increased card height for layout.
  * App bar styling flattened by removing scrolled-under elevation.

### DIFF
--- a/lib/app.dart
+++ b/lib/app.dart
@@ -26,8 +26,6 @@ class _MyAppState extends State<MyApp> {
   Widget build(BuildContext context) {
     _setup(context);
 
-    Stores.setting.useSystemPrimaryColor.fetch();
-
     return ListenableBuilder(
       listenable: RNodes.app,
       builder: (context, _) {
--- a/lib/view/page/setting/entries/app.dart
+++ b/lib/view/page/setting/entries/app.dart
@@ -104,7 +104,15 @@ extension _App on _AppSettingsPageState {
                     DynamicColorBuilder(
                       builder: (light, dark) {
                         final supported = light != null || dark != null;
-                        if (!supported) return const SizedBox.shrink();
+                        if (!supported) {
+                          if (!_setting.useSystemPrimaryColor.fetch()) {
+                            _setting.useSystemPrimaryColor.put(false);
+                            WidgetsBinding.instance.addPostFrameCallback((_) {
+                              setState(() {});
+                            });
+                          }
+                          return const SizedBox.shrink();
+                        }
                         return ListTile(
                           title: Text(l10n.followSystem),
                           trailing: StoreSwitch(
@@ -141,9 +149,14 @@ extension _App on _AppSettingsPageState {
       return;
     }
 
-    UIs.primaryColor = color;
-    UIs.colorSeed = color;
+    // Save the color seed to settings
     _setting.colorSeed.put(color.value255);
+
+    // Only update UIs colors if we're not in system mode
+    if (!_setting.useSystemPrimaryColor.fetch()) {
+      UIs.primaryColor = color;
+      UIs.colorSeed = color;
+    }
 
     RNodes.app.notify();
     context.pop();


GT-610:
> **Actionable comments posted: 0**
> 
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > lib/app.dart (1)> `65-93`: **Potential color inconsistency when dynamic colors are unavailable.**
> > > If both `light` and `dark` are `null` (on platforms that don't support dynamic colors), the theme will use Flutter's default color scheme, but `UIs.primaryColor` and `UIs.colorSeed` won't be updated. This could cause inconsistency between the applied theme and UI elements that directly reference `UIs.primaryColor`.
> > > Consider adding a fallback to update `UIs` values when dynamic colors are unavailable:
> > > 
> > > 🔧 Suggested fix
> > > ```diff
> > >          if (context.isDark && dark != null) {
> > >            UIs.primaryColor = dark.primary;
> > >            UIs.colorSeed = dark.primary;
> > >          } else if (!context.isDark && light != null) {
> > >            UIs.primaryColor = light.primary;
> > >            UIs.colorSeed = light.primary;
> > > +        } else {
> > > +          // Fallback when dynamic colors unavailable
> > > +          final fallback = Color(Stores.setting.colorSeed.fetch());
> > > +          UIs.primaryColor = fallback;
> > > +          UIs.colorSeed = fallback;
> > >          }
> > > ```
> 
> 🧹 Nitpick comments (2)
> > lib/app.dart (1)> `28-29`: **Unused fetch result appears to be dead code.**
> > > The `fetch()` call on line 29 is executed but its return value is discarded. If this is intended for cache warming, consider adding a comment. Otherwise, this line can be removed since `fetch()` is called again inside the `ListenableBuilder` on line 34.
> > > 
> > > 🔧 Suggested fix
> > > ```diff
> > >    _setup(context);
> > > -
> > > -   Stores.setting.useSystemPrimaryColor.fetch();
> > > 
> > >    return ListenableBuilder(
> > > ```
> > 
> > 
> > 
> > lib/view/page/setting/entries/app.dart (1)> `131-144`: **Consider also updating `UIs.primaryColor` for consistency.**
> > > The `_onSaveColor` method updates `UIs.colorSeed` but not `UIs.primaryColor`. In `lib/app.dart`, both values are updated together. While the subsequent `RNodes.app.notify()` will trigger a rebuild that updates both, there's a brief window where these values could be inconsistent.
> > > 
> > > 🔧 Suggested fix
> > > ```diff
> > >    void _onSaveColor(String s) {
> > >      final color = s.fromColorHex;
> > > 
> > >      if (color == null) {
> > >        context.showSnackBar(libL10n.fail);
> > >        return;
> > >      }
> > > 
> > >      UIs.colorSeed = color;
> > > +    UIs.primaryColor = color;
> > >      _setting.colorSeed.put(color.value255);
> > > 
> > >      RNodes.app.notify();
> > >      context.pop();
> > >    }
> > > ```
> 
> 📜 Review details

@coderabbitai Noticed this. This might impact systems without dynamic color or light/dark switch (Android 9-, Windows, etc.) I think, but I don't know if it really matters.

But it won't cause any bugs. Investigating.

GT-610:
Screenshot_1768046742

I added a logic that directly hide "Follow system" feature when the device does not support this feature.

This screenshot is taken from Android 9, which does not have dynamic color support.

@coderabbitai Please review.

GT-610:
Unused call problem is being fixed.

PR #1019, opened 2026-01-22
fix(sftp): FIx Permission Denied While Accessing SFTP

### LINKED ISSUE #1012: [Bugs + Feature Requests] `Permission Denied` While Accessing SFTP.
**Describe the bug**
When Trying To Access Files Using SFTP, It Shows `Permission Denied` Message. SFTP Server Works Perfectly. The Problem Is The Server Does Not Have Access To The Path, The Client Side Is Trying To Access, Which Is `/`. 

**To Reproduce**
1. Have A Normal Connection To Server. 
2. Open File Explorer In App. 
3. See Error As The App Tries To Open `/` While The User Have No Access To It. 

**Desired Results**
1. Editable Path Displaying Text, Which After Change And Enter Key Pressed, Switches To That Directory. 
2. Option To Set Default Path In Settings. Follows `/` If Blank. 

These Should Be Able To Fix The Problem. 

**Actual Results**
Access Denied. 

**Additional context**
SFTP Commands May Work In Operations Where Access Is Possible, Regardless Of Access Of Default Path Is Not Possible. 

Example :- If User Has Access To Path `/home/usr1/` , And No Access At `/` , Operations Can Work, If Each Operation Contains The Accessible Path As Prefix. Like For Create Folder :- `/home/usr1/tempdir` Will Still Work, Even If The User Is At `/` .

GT-610:
Resolve #1012.

## Summary by CodeRabbit

* **Chores**
  * Added six new contributors to the participants list.

* **Bug Fixes**
  * SFTP connections now start in each user's home directory (or /root for root), improving the initial location for file browsing.
  * Initial remote directory listing begins asynchronously to avoid delaying the UI.
  * When enabled, previously opened path history can override the initial path.

### DIFF
--- a/lib/view/page/storage/sftp.dart
+++ b/lib/view/page/storage/sftp.dart
@@ -85,7 +85,7 @@ class _SftpPageState extends ConsumerState<SftpPage> with AfterLayoutMixin {
     try {
       final homeResult = await _client.run('eval echo ~${widget.args.spi.user}');
       final homePath = homeResult.string.trim();
-      if (homePath.isNotEmpty && homePath != '~') {
+      if (homePath.isNotEmpty && homePath.startsWith('/')) {
         initPath = homePath;
       } else {
         final user = widget.args.spi.user;


GT-610:
@coderabbitai In all future pull requests (PRs), please add all participants from the issues resolved by that PR to the `participants` field in `lib/data/res/github_id.dart`. Please first check whether the `participants` in the issues associated with this PR already exist or have been added in the commit. If not, please propose them.

PR #1238, opened 2026-07-20
fix: use systemctl to list units

Sandarr95:
On NixOS the systemd tab is empty, as listing unit files is done through `find -type f ...`, but on NixOS these are symlinks to the nix store. Using `systemctl` avoids any discrepancies and ensures future systemd changes are followed. A specific error for missing user units or no systemd connection is also added.

## Summary by CodeRabbit

* **New Features**
  * Enhanced system and user unit discovery.
  * Added robust parsing for `systemctl list-units` output, including unit type, scope, state, and multi-word descriptions.
  * Introduced refresh status reporting for system vs user loading failures.

* **Bug Fixes**
  * Improved resilience to incomplete, unsupported, or malformed unit output.
  * Refresh behavior is now more consistent, with proper completion handling and user notifications.

* **Tests**
  * Added unit parsing tests covering valid output, empty/bus-error cases, and invalid column structures.

### DIFF
--- a/lib/data/model/server/systemd.dart
+++ b/lib/data/model/server/systemd.dart
@@ -42,6 +42,15 @@ enum SystemdUnitScope {
     }
     return 'systemctl --user';
   }
+
+  /// Command to enumerate units for this scope.
+  ///
+  /// Listing is read-only, so no `sudo` is needed even for the system scope.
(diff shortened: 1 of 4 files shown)


GT-610:
@coderabbitai review

Issue #1262, opened 2026-08-09
feat: reuse local ssh-agent socket and ControlMaster connections on desktop


lollipopkit:
## Motivation

On desktop, the app cannot reuse anything from the SSH setup the user already has on the same machine.

Key auth today requires importing the private key **into** the app: `PrivateKeyInfo.key` holds the PEM as plaintext in Hive, and `getPrivateKey()` (`lib/core/utils/server.dart`) reads it back by id. That rules out every setup where the key material is not extractable:

- hardware-backed keys (YubiKey, Secure Enclave via Secretive)
- password-manager agents (1Password, KeePassXC)
- `gpg-agent`, or keys held by the system keychain
- environments where exporting a private key is against policy

It also means a user with a working `ssh` on the same box has to duplicate — and separately secure — their credentials inside the app.

Two separate things should be reusable from the host. They share a theme but almost no implementation.

---

## Part 1 — ssh-agent authentication

Sign the `publickey` auth challenge with the host's running agent instead of a locally stored PEM. The app never sees the private key.

### Current state

`genClient()` (`lib/core/utils/server.dart:235-260`) offers exactly three auth paths: password, keyboard-interactive, and an in-app PEM. There is no way to reach an external agent.

The vendored fork already contains the agent protocol's **server** side — `SSHKeyPairAgent` / `SSHAgentChannel` in `packages/dartssh2/lib/src/ssh_agent.dart` — which answers `SSH_AGENTC_REQUEST_IDENTITIES` / `SSH_AGENTC_SIGN_REQUEST` from in-app keys, for forwarding to a remote. It is not wired into `lib/` at all (`agentHandler` has zero references there). What is missing is the **client** side: connect to the host agent and ask *it* to sign.

### Blocker: `SSHKeyPair.sign()` is synchronous

`packages/dartssh2/lib/src/ssh_key_pair.dart:66` declares `SSHSignature sign(Uint8List data)`, and both call sites are in synchronous `void` methods:

- `ssh_client.dart:1505` — `_authWithNextPublicKey()`
- `ssh_client.dart:1547` — `_authWithNextHostbased()`

Agent signing is async IPC over a socket, so the fork has to grow an async signing path. Suggested shape, kept additive so existing key types are untouched:

```dart
abstract class SSHSigner {
  String get type;
  SSHHostKey toPublicKey();
  Future<SSHSignature> sign(Uint8List data);
}
```

[...]

PR #1312, opened 2026-08-18
fix(ssh): an imported IdentityFile is a path, not a key store id

### LINKED ISSUE #1263: fix: keyId is overloaded — SSH config import writes an IdentityFile path where a key store id is expected
## Summary

`SshCredential.keyId` has two incompatible meanings depending on who wrote it. SSH config import writes a filesystem **path**; everything that consumes it expects a private-key store **id**. Servers imported from `~/.ssh/config` with an `IdentityFile` therefore fail to connect.

## Details

`SSHConfig._parseSSHConfig` stores the raw `IdentityFile` value straight into `keyId` — `lib/core/utils/ssh_config.dart:163-165` and `:91`:

```dart
case 'identityfile':
  identityFile = value; // Store the path directly
  ...
ssh: SshCredential(..., keyId: identityFile, ...)
```

The consumer treats it as a key id — `lib/core/utils/server.dart:41-50`:

```dart
String getPrivateKey(String id) {
  final pki = Stores.key.fetchOne(id);
  if (pki == null) {
    throw SSHErr(type: SSHErrType.noPrivateKey, ...);
  }
  return pki.key;
}
```

Nothing in between translates a path into a store entry — the parsed `Spi`s go from `SSHConfig.parseConfig()` through dedup straight into `Stores.server` via `ServerDeduplication.importServersWithNotification` (`lib/view/page/setting/entries/ssh.dart:155-220`).

The edit page has the same mismatch: it resolves the key selector by looking `keyId` up among stored keys (`lib/view/page/server/edit/actions.dart:486`, `:495-498` — `indexWhere((e) => e.id == ssh.keyId)`), so an imported path matches nothing and the key field silently shows empty.

## Reproduction

1. On desktop, have a `~/.ssh/config` entry with an `IdentityFile` line.
2. Settings → import SSH config, accept the import.
3. Connect to the imported server.

Expected: connects using that key, or a clear message that the key must be imported first.
Actual: `SSHErr(noPrivateKey)` with the file path in the message. Opening the server in the edit page shows no key selected.

## Notes on the fix

Deciding between the options is the substance here:

- Read the `IdentityFile` at import time and create a `PrivateKeyInfo` in `Stores.key`, then store its id. Copies key material into the app, and breaks on encrypted or hardware-backed keys.
- Keep the path as a distinct field (e.g. `keyPath` next to `keyId`) and load it at connect time. Needs a migration and file-access permission on macOS.
- Import the server without a key and surface a "this host needs a key" hint.

Whichever way it goes, `keyId` should stop carrying two meanings.

Related: #1262 adds ssh-agent auth, which is the actual answer for most `IdentityFile` setups — the agent already holds the key. That issue will touch this field anyway, but this bug stands on its own and is worth fixing first.

lollipopkit:
Fixes #1263.

## The bug

A server imported from `~/.ssh/config` with an `IdentityFile` could never connect. `SSHConfig` wrote the path into `SshCredential.keyId` (`ssh_config.dart:91`), and every reader looks that field up in `Stores.key` (`server.dart:40-49`) — so the connection failed with `noPrivateKey` naming a path as a missing id, and the edit page's key picker, which matches stored ids (`edit/actions.dart:494-498`), showed nothing selected.

Three assertions in `ssh_config_test.dart` expected `keyId` to hold the path. They encoded the bug, which is a large part of why it shipped.

## The fix

`keyId` means a `Stores.key` id and only that. The path gets `keyPath`, read when the connection is made rather than copied into the app at import time — an encrypted or hardware-backed key stays where the user put it, which is also the direction #1262 needs.

`resolvePrivateKey` is now the one place that turns a credential into a PEM. Both callers go through it:

- `genClient`
- `SshTransferCreds.forServer`, which runs on the main isolate because the transfer isolate has no stores, no granted filesystem, and nobody to report a refusal to. Its `privateKeysByKeyId` map is keyed by `SshCredential.keyRef` now, since a key that came from a path has no id to be keyed by.

A jump server whose key cannot be resolved is logged and skipped rather than fatal — that hop may authenticate by password, and one bad candidate must not take the others with it. The target server's key still throws.

## Migration

Records written before this move on startup, next to `migrateIds` in `main.dart`.

Decided against the key store, not by the shape of the string: a key's id is its **user-typed name** (`private_key/edit.dart:282`), so somebody may legitimately have called one `~/keys/prod`. A value that really is an id always wins; only a path-shaped value that names no key moves.

The Hive adapter change is additive — `keyPath` takes a new field index, so a record written before it reads back as null.

## Edit page

Shows the file and where it came from, and carries it through a save. Building the credential from the form alone would have silently dropped the only credential an imported server has, the first time someone opened and saved one. Picking a stored key clears the path; the two are mutually exclusive.

## Errors

Failures name the file and the reason. On the App Store build they name the sandbox instead: `~/.ssh` is outside its container, so that path cannot work there at all, and the file error it would otherwise produce says nothing about why. Two new strings, in all 15 locales.

## Verification

- `flutter test` — 896 pass
- `flutter analyze lib test integration_test` — clean
- New `test/identity_file_key_test.dart`: `keyRef` disambiguation, JSON round trip, the migration's four cases (moves a path, leaves a path-shaped *id* alone, leaves an ordinary id alone, idempotent), and `resolvePrivateKey` reading a real file / naming a missing one
- `ssh_config_test.dart` updated to assert `keyPath`

Incidentally regenerated: `single.g.dart`'s riverpod hash was stale on main.

## Summary by CodeRabbit

- **New Features**
  - SSH connections now support private keys selected from storage or imported from filesystem paths.
  - Imported SSH key paths are displayed in server settings and can be cleared or replaced.
  - Home-directory paths (`~`) are supported in SSH configurations.
  - Existing SSH identity-file settings are migrated automatically.

- **Bug Fixes**
  - Improved handling and localized error messages for missing, unreadable, or sandbox-restricted private-key files.

### Changes

- **SSH credential identity-file contract and connection-time key resolution**: Adds a distinct filesystem key-path field and key-reference namespace, resolves imported private-key files at connection/transfer setup, and updates SSH configuration parsing to populate the new field.
- **Persisted server migration and startup ordering**: Migrates legacy IdentityFile values incorrectly stored in keyId into keyPath when they are not actual private-key IDs, and invokes that migration during application initialization.
- **SSH config parsing and first-run import behavior**: Updates SSH config import to retain IdentityFile paths, expand home paths when opening them, normalize ProxyJump/ProxyCommand precedence, and adds parser coverage.
- **Server editor credential UI and save semantics**: Adds display/preservation/clearing of imported key paths, keeps stored-key selection mutually exclusive with file credentials, and maintains SSH-versus-monitor editing behavior.
- **Localization and generated model/provider artifacts**: Adds localized private-key-file and sandbox messages across source ARB files and regenerated localization outputs, plus regenerated Hive/Riverpod artifacts reflecting the model changes.


lollipopkit:
Two of the three taken, one skipped.

**`keyRef` namespaces — fixed.** The finding is right and my justification for the old shape was wrong: the comment claimed "a store id never contains a separator", but a key's id is its *user-typed name* (`private_key/edit.dart:282`), so it can be `path:/home/me/id_ed25519` as easily as anything else. Both sides are prefixed now. The value is never persisted — `SshTransferCreds` has no `toJson` and lives only as long as the transfer it was built for — so the shape was free to change. Test added for the collision case.

**Key auth ignoring `keyPath` — fixed, and it was worse than the comment says.** Both call sites read only the stored-key selection, so a server imported with an `IdentityFile` had a key and no password and therefore:

- the switch showed *off* while key authentication was on, and
- every save asked "continue without authentication?" about something that was not true.

Both read either source now, and turning the switch off clears both. Turning it on when a file is already set leaves the file alone rather than adding an empty stored-key selection beside it.

While in there: the save path had a related hole the review did not name — `_onSave` rebuilds the credential from the form, and nothing on the form types a path, so opening an imported server and saving it once removed the only credential it had. That is now carried through, with a widget test that fails (`keyPath` null) when the carry-through is removed.

**Jump failover on `SSHErrType.noPrivateKey` — skipped.** It is pre-existing behaviour rather than something this branch introduced: the same path called `getPrivateKey(keyId)` before, which throws the same error, so a jump server with an unresolvable key already aborted the chain. Widening `_isJumpFailoverError` from network faults to configuration errors would also change what happens when a jump server's *stored* key has been deleted — it would silently route around it via another candidate instead of saying so. That is a product decision about what failover means, and it belongs in its own change rather than in a fix for the `keyId` overload.

Note the asymmetry is deliberate on the transfer side and documented there: `SshTransferCreds._tryResolve` does skip a jump server whose key cannot be resolved, because that runs at *queue* time across all candidates, where failing the whole transfer over candidate #2 would be wrong.

`flutter test` 898 pass, `flutter analyze lib test integration_test` clean.

lollipopkit:
The `winnowl` review is against `5413435e`, the first commit here — its `keyRef` collision finding and CodeRabbit's key-auth one were already fixed in `afaa0b43`, before it ran. Two findings were new.

**Socket leak — fixed.** `genClient` opens the socket first and resolves the key after, so every throw between the two returned a key error and no handle: nothing the caller had could close what was already open. A status poll retries, so that is one leaked connection per attempt, or on the `ProxyCommand` path one leaked `/bin/sh` per attempt.

Pre-existing — `getPrivateKey` threw from the same place — but widened here, since a `keyPath` adds three ways to fail where there was one.

Client construction moves into `_authenticatedClient` so `genClient` has a single place to undo the socket from. `destroy()` rather than `close()`: nothing is buffered worth flushing, and `destroy` is what kills the `ProxyCommand` process.

Has a test, against a real loopback listener that has to observe the close. Removing the `destroy` makes it time out waiting for one.

Not addressed, deliberately: the review also notes a successful jump path retains its forwarding client. That is true and older than this branch — `genClient` returns a client whose socket is a forward over a jump client nobody tracks. It needs an owner for that client, which is a change to the shape of this function rather than to this fix.

**Bare relative `IdentityFile` not migrated — skipped, and written down.** Correct that `IdentityFile id_ed25519` is legal and that `_looksLikePath` will not move it.

It is also indistinguishable from a stored key named `id_ed25519`, and moving it would need a rule for what a bare name resolves against. OpenSSH's answer to that is not one to guess at, and guessing wrong turns a record that fails cleanly into one that opens the wrong file or a path relative to whatever the app's working directory happens to be. If it is worth handling, it belongs at the parse site — on values not yet stored — rather than in a migration over ones that are. The gap is now in `migrateIdentityFilePaths`'s doc comment.

**Jump failover on `noPrivateKey`** — still skipped, for the reason given earlier.

`flutter test` 899 pass, `flutter analyze lib test integration_test` clean.

---

Unrelated, found while grepping this file and worth its own change: `lib/core/utils/server.dart` contains a literal NUL byte (`'${info.keyType}\x00${info.fingerprintHex}'`, in `promptHostKeyExclusively`). It predates this branch — `origin/main` has it. It works, but `rg` classifies the file as binary and skips it, so every search across `lib/` silently misses this file. `String.fromCharCode(0)` would keep the separator and make the source text again.

PR #1313, opened 2026-08-18
ci(macos): trim the build triggers; drop the committed coverage report

lollipopkit:
`macOS build` ran on #1312 and could not have told it anything. It fires on `lib/main.dart`, and that PR adds one line there.

## Why those paths were in the list

They are left over from when this workflow ran `integration_test` on a macOS device — `main.dart`, `sandbox_import.dart`, `local_shell.dart` and `integration_test/**` were that suite's subject. It never passed on a hosted runner (the file's own header documents why) and was changed to build instead. The trigger list was not.

`flutter build macos` does not compile `integration_test/` at all, so that entry could never have mattered even then.

## What the list should be

What only a macOS runner can answer: whether the project, the pods and the Rust static library still link. A Dart change that would fail to compile fails in `analysis.yml` first — `flutter analyze lib test integration_test` plus `flutter test` — minutes earlier and on a cheaper runner.

| | why |
| --- | --- |
| `macos/**` | the Xcode project, the entitlements, the Podfile |
| `crates/**`, `Cargo.toml`, `Cargo.lock` | cargokit links the Rust library into the app; nothing else in CI links it on macOS |
| `pubspec.yaml`, `pubspec.lock` | a plugin added or moved changes what `pod install` resolves and links |
| `packages/**` | widened from `packages/fl_lib` — every vendored package is a submodule, so a gitlink move is a source change to what gets compiled |
| `.github/workflows/macos.yml` | the filter is what is under test then |

Dropped: `lib/main.dart`, `lib/core/utils/sandbox_import.dart`, `lib/core/utils/local_shell.dart`, `integration_test/**`.

## Check

Against #1312's actual file list the new filter matches nothing — that PR would not have started this job. `actionlint` clean.

This workflow keeps its own `paths` filters rather than moving to the `changes` job added in #1310: that job exists because `analysis.yml` has several jobs with different scopes, and a single-job workflow has nothing to gate against itself.

## Summary by CodeRabbit

* **Chores**
  * Updated macOS automation to build the macOS app.
  * Refined change detection for relevant macOS, Rust, Flutter package, vendored dependency, or workflow updates.
  * Removed unnecessary integration test and unrelated Dart file triggers.
  * Added coverage output to ignored files.

* **Documentation**
  * Clarified device integration test organization and analysis coverage.
  * Added wireless iOS 17+ testing instructions and updated the integration test command.

* **Testing**
  * Added a standard integration test driver for running device tests.

### Changes

- **macOS CI workflow: native build coverage and local signing**: The macOS GitHub Actions job is converted from attempted device integration testing to an ad-hoc/debug macOS build, with narrowed path triggers, dependency/toolchain setup, and an in-job project/entitlements rewrite.
- **Developer testing documentation and integration-test boundaries**: The testing guide is expanded to document Rust/FFI commands, opt-in SSH and monitor tests, frontend checks, widget/provider examples, deterministic dependency rules, and the distinction between unit, integration, and networked-device execution.
- **Coverage artifact and ignore-rule hygiene**: The previously committed coverage report is removed and the repository ignore configuration is adjusted so generated coverage output is treated as local build/test output rather than source.


lollipopkit:
Folded in two related bits of repo hygiene rather than opening more PRs.

**`coverage/lcov.info` — removed and gitignored.** Committed in #836, last written a year ago, and nothing reads it: no codecov config, no CI step, no badge. It still names `lib/data/model/app/shell_func.dart`, which has since moved to `lib/data/model/app/scripts/`. `make coverage` (`flutter test --coverage`) regenerates it on demand, so it was a build output tracked by accident.

**`test_driver/integration_test.dart` — kept, and now documented.** It looks like dead weight until you open it. It exists for exactly one case: an iOS 17+ device Xcode reaches over the network, where `flutter test` cannot launch the app at all — it hardcodes `disablePortPublication: true` and `IOSDevice.startApp` refuses a wirelessly tethered device when that is set, with no flag to clear it. `flutter drive --publish-port` is the same run with the bit cleared. That reason existed only in the file's own doc comment; it is in `docs/development/testing.md` now, with the command.

Same section also picked up why `integration_test/` is its own directory: `flutter test` with no argument runs everything under `test/`, so device tests placed there would be picked up by every local run and by CI, which has no device.

Docs site builds (53 pages).

lollipopkit:
Both taken.

**YAML anchor.** Right, and the comment I inherited (`Duplicated rather than anchored: GitHub Actions does not resolve YAML aliases`) predated the September 2025 change. One list now, aliased into `pull_request`. `actionlint` 1.7.12 clean. Worth noting the duplication had already done what duplication does: the two halves were identical here only because nobody had edited one of them.

Live check for the alias actually resolving on GitHub's side, not just in `actionlint`: this PR modifies `.github/workflows/macos.yml`, which is in the list. If `macOS build` starts on this push, `pull_request.paths` resolved to a non-empty list.

**Docs sentence.** Taken — it was a garden path. Condition first now.

---

Separately, on `test_driver/`: **not deleting it**, and I want to flag why before anyone does.

I had it as "exists for one narrow case", but it has a second, concrete user I missed. `integration_test/ios_bench_test.dart:21` documents the only way to run the iOS benchmark:

```
flutter drive --driver=test_driver/integration_test.dart \
  --target=integration_test/ios_bench_test.dart \
  -d <device> --profile --publish-port \
  --dart-define-from-file=/tmp/bench.json
```

`flutter test` cannot stand in for that: it has no `--profile`, and a benchmark run in debug mode measures the wrong thing. So deleting the driver would take the benchmark with it, not just the wireless-device path.

PR #1234, opened 2026-07-18
monorepo finale: relocate crates, monitor release pipeline, docs overhaul

lollipopkit:
## Summary

This PR finishes the monorepo migration with three main blocks, plus follow-up hardening:

### 1. migrate: merge `rust/` and `rust_builder/` into `crates/sbm_ffi`
- The cargokit plugin shell and the sbm_ffi crate now share one directory (cargokit's official example layout), keeping the top level clean
- Workspace members, FRB `rust_root`, pubspec paths, and per-platform glue updated in sync
- Verified: `cargo test --workspace` green, 12/12 FFI parity tests pass, `flutter build macos --debug` builds through cargokit for real

### 2. feat: monitor release pipeline
- Manual trigger only; version comes from the monitor crate (`monitor-v*` tags), decoupled from the app's `v1.0.x`
- Native binaries: Linux musl (amd64/arm64, native arm runners) / macOS (both archs) / Windows → GitHub Release + SHA256SUMS
- Docker: the Linux musl binary is reused in an Alpine image pushed to ghcr.io — zero in-image compilation, no QEMU
- Drive-by fixes: Dockerfile frontend path mismatch with server.rs, build/runtime glibc mismatch, reqwest switched to rustls to drop the OpenSSL dependency (prerequisite for static musl builds)
- `install.sh` rewritten for the new release layout; fresh installs generate a random `JWT_SECRET`
- Verified: local image build + run smoke test (health 200, frontend OK, uptime collection OK), 41 monitor tests green

### 3. docs: full audit of the docs site (en/zh/ja/de/fr/es)
- Removed content confirmed fabricated after checking against the code: system tray, Touch Bar, multi-window, NUL keep-alive, SFTP paging/TTL cache, nonexistent keyboard settings, etc.
- Fixed factual errors: home tab list, Windows integration language, encrypted-storage description, class/setting names, nonexistent test file examples
- Added monorepo layout, Rust testing, and FRB codegen docs; README monitor links point into the monorepo
- Added dependabot config (cargo × 3 directories + monitor frontend npm + GitHub Actions)
- Removed doc/adr (intentional) and cleaned up all references
- Verified: docs site builds, 139 pages

### 4. Follow-ups on this branch
[...]

PR #1315, opened 2026-08-18
fix(macos): a sandboxed ProxyCommand failure now says why

### LINKED ISSUE #1264: fix: ProxyCommand likely broken on the sandboxed macOS build
## Summary

The macOS build enables App Sandbox, which very likely breaks `ProxyCommand` — the feature spawns `/bin/sh`, and the child inherits the sandbox, so any proxy command that reads `~/.ssh/` is denied.

**Not yet verified on a real signed build** — filing so it gets checked rather than assumed working.

## Details

`macos/Runner/DebugProfile.entitlements` and `Release.entitlements` both set:

```xml
<key>com.apple.security.app-sandbox</key>
<true/>
```

The granted exceptions are `network.client`, `network.server`, `files.user-selected.read-write`, and iCloud containers. There is no access to `~/.ssh/`.

`ProxyCommandSocket.connect` (`lib/core/utils/proxy_command_socket.dart:50-54`) does:

```dart
final process = await Process.start(
  shellCommand.executable,   // '/bin/sh'
  shellCommand.arguments,    // ['-c', command]
  mode: ProcessStartMode.normal,
);
```

The sandbox does not block `fork`/`exec` itself, but the child runs under the same profile. So the expected outcome splits by what the command actually is:

- `ProxyCommand ssh -W %h:%p jump` — the child `ssh` reads `~/.ssh/config` and `~/.ssh/id_*`, both outside the container. Should fail.
- `ProxyCommand cloudflared access ssh ...` — reads `~/.cloudflared/`. Should fail.
- `ProxyCommand nc %h %p` — touches no home-directory path and only needs network access, which is granted. May well work.

If that reading is right, the feature works for a narrow set of commands and fails for the common ones, and the failure surfaces only as `ProxyCommand exited with code N` plus whatever the child wrote to stderr (`proxy_command_socket.dart:85-97`) — there is nothing pointing at the sandbox as the cause.

## To verify

Build a sandboxed release (`dart run fl_build -p macos`), configure a server with `ProxyCommand ssh -W %h:%p <somehost>`, connect, and check `Console.app` for sandbox denials from the child process.

## Options

- Drop `com.apple.security.app-sandbox` from the directly distributed DMG / Homebrew cask build, and gate the feature off on a Mac App Store build. This is the direction already chosen for ssh-agent support in #1262, and the two would share one capability flag.
- Keep the sandbox and detect the situation, so the error names the cause instead of surfacing a bare exit code.

Either way the error message deserves improvement — right now a sandbox denial is indistinguishable from a genuine command failure.

Related: #1262 (ssh-agent / ControlMaster reuse) hits the same wall for the same reason, and proposes the same capability flag.

lollipopkit:
Fixes #1264.

## Verified, and it is worse than the issue predicted

The issue asked for this to be checked rather than assumed. It cannot be checked the obvious way: `flutter build macos --release` signs with the **App Store distribution profile** (no `ProvisionedDevices`, no `get-task-allow`), which launchd refuses to start locally — `Launchd job spawn failed`, POSIX 163. Re-signing it ad hoc is refused too, by AMFI: `The file is adhoc signed but contains restricted entitlements`, `app-sandbox` being one.

So: one small binary, signed twice with a Development identity — once with `app-sandbox`, once without — spawning `/bin/sh -c` children exactly as `ProxyCommandSocket` does.

| probe | sandboxed | not sandboxed |
| --- | --- | --- |
| `$HOME` as the child sees it | `~/Library/Containers/<id>/Data` | `/Users/<me>` |
| `$HOME/.ssh/config` | no such file | reads it |
| `/Users/<me>/.ssh/config` | `Operation not permitted` | reads it |
| `ls /Users/<me>/.ssh` | `Operation not permitted` | lists it |
| `ssh -W 127.0.0.1:22 <alias>` | exit 255 | exit 0, SSH banner |
| `/dev/tcp/127.0.0.1/22` | works | works |

The issue predicted "the child `ssh` reads `~/.ssh/config` ... should fail". True, but the mechanism is not a denial — **`$HOME` is replaced**, so `~` silently points at the container. Only an absolute path into the real home reports `Operation not permitted`.

Which makes the failure unrecognisable. `ssh -W %h:%p somealias` never reads `~/.ssh/config`, so the alias is never resolved; ssh takes it for a literal hostname and reports:

```
ssh: connect to host somealias port 22: Operation timed out
```

A network fault, naming the user's jump host. The app passed that through as `ProxyCommand exited with code 255.` — nothing anywhere said sandbox.

The issue's other prediction also holds: `nc %h %p` works. Network access is granted, and `/dev/tcp` confirms it.

## The fix

Not a refusal. That last row is real — a command touching no home-directory path works on the sandboxed build, and blocking the field would take that away.

- Every failure path in `ProxyCommandSocket` appends what the sandbox does, **appended rather than substituted**: the command's own outcome is the only thing naming which command and how it ended.
- The edit page says it beside the field, so it is visible before the failure rather than only after.
- The wording avoids "denied" on purpose. The child gets a *different* home, so what it reports is a missing file or an unreachable host — someone told "permission denied" would go looking for a dialog that does not exist.

Uses `Pfs.isMacSandboxed`, the same runtime check `LocalShellBackend.isSupported` already uses, so one binary is honest in both distributions. Nothing new was needed: the DMG build has had `app-sandbox = false` since #1234.

## Verification

`test/proxy_command_sandbox_test.dart` locks the shape both ways — appended not substituted, silent when unconfined, and naming the home directory rather than permissions. `flutter test` 896 pass, `flutter analyze lib test integration_test` clean.

## Summary by CodeRabbit

* **New Features**
  * Added clear guidance for ProxyCommand failures in macOS sandboxed builds.
  * Added an explanatory note beneath the ProxyCommand field when sandbox restrictions apply.
  * Network-only commands remain supported, while commands accessing `~/.ssh` may require the unsandboxed DMG version.

* **Localization**
  * Added translated sandbox guidance across supported languages.

* **Bug Fixes**
  * Improved connection, timeout, and process-exit error messages with actionable explanations.

### Changes

- **ProxyCommand sandbox-aware diagnostics and socket lifecycle**: Adds sandbox-specific explanatory text to ProxyCommand connection failures and exposes the explanation helper for tests, while retaining the process-backed SSHSocket implementation.
- **Server editor ProxyCommand presentation and save contract**: Adds a sandbox warning alongside the ProxyCommand editor on confined macOS builds and integrates the existing localized warning into the server-edit experience.
- **Localization source and generated ProxyCommand sandbox contract**: Introduces the proxyCommandSandboxed message in the English and translated ARB catalogs and regenerates the localization interface and locale implementations.


lollipopkit:
Both taken.

**German.** `statt Ihres` did leave the genitive noun implied. Reordered rather than appended, though — `statt Ihres eigenen Home-Verzeichnisses` names the same noun twice in one clause, since the sentence already says `ein leeres Home-Verzeichnis`. It now reads:

> Der Befehl sieht statt Ihres Home-Verzeichnisses ein leeres

**French.** Right, `expiration` is not the term. Now `sous la forme d'un délai d'attente`.

Checked before changing either: neither file had a prior translation of "timeout" to be consistent with — this string is the first use in both.

`flutter test` and `flutter analyze lib test integration_test` still clean.

Issue #1269, opened 2026-08-10
feat: out-of-band management (BMC) via Redfish, with IPMI as an optional fallback


lollipopkit:
## Summary

Add out-of-band management for servers that have a BMC, so that power state,
hardware sensors and hardware event logs are reachable while the host OS is
down, hung or otherwise unreachable.

## Protocol choice: Redfish or IPMI

Recommendation: implement Redfish first, and keep IPMI over LAN as an optional
later fallback.

| Aspect | Redfish | IPMI 2.0 over LAN |
| --- | --- | --- |
| Transport | HTTPS + JSON (REST) | RMCP+ over UDP 623, binary |
| Client work in this repo | `dio`, no native code | no Dart implementation exists; needs a client in `crates/` behind FFI. `ipmitool` cannot be invoked on iOS/Android |
| Data model | self-describing: `Systems`, `Chassis`, `Thermal`/`Power`, `Storage`, `LogServices`, `Managers` | SDR / SEL / chassis commands; anything beyond that is vendor-specific `raw` commands |
| Auth | TLS, session token via `POST /redfish/v1/SessionService/Sessions` | RAKP handshake, with protocol-level weaknesses that firmware cannot fix (password hash retrieval for any known username, CVE-2013-4786; cipher suite 0 auth bypass) |
| Hardware coverage | most enterprise hardware from around 2016 on: iDRAC 8/9, iLO 5/6 (iLO 4 partial), Lenovo XCC, Supermicro X11+, ASRock Rack, Huawei iBMC, OpenBMC | everything above, plus older and entry-level BMCs that predate Redfish |
| Spec status | active DMTF spec | last revision is 2.0 rev 1.1 (2013); DMTF positions Redfish as the successor |
| Serial console | not covered; vendors expose HTML5 KVM instead | SOL, which maps onto the existing xterm |

Redfish covers power, inventory and sensors at a fraction of the implementation
cost here. IPMI stays relevant only for pre-Redfish hardware and for SOL.

## How it fits the current architecture

- A BMC is a separate endpoint from the host. It does not belong on the
  `ServerConnectCredential` axis, which answers where the host's *status* comes
  from. Model it like `WolCfg`: an optional side channel on `Spi`
  (address, credential, certificate trust), independent of how the host itself
  is reached.
- `ServerCapabilities` gains flags for what the BMC provides (power control,
  hardware sensors, event log), so the UI keeps asking capabilities instead of
  testing which protocol is configured.
[...]

PR #1316, opened 2026-08-18
feat(bmc): out-of-band management over Redfish (Phase 1)

lollipopkit:
Closes part of #1269. Redfish only; IPMI is not here and is not planned for this PR — see the end.

The PR started as documentation and has since grown the implementation. What follows describes what is on the branch now.

## What a BMC is, in this app's terms

A side channel on `Spi`, beside `wolCfg`. Not a third `ServerConnectCredential`: that axis answers "where does this server's *status* come from" and is dispatched per transport, while a BMC answers about the machine when the host is absent. The two are also not exclusive — the normal case is SSH *and* a BMC on one record, which the SSH-xor-monitor `CHECK` would have refused.

```dart
class Spi {
  SshCredential? ssh;
  MonitorHttpCredential? monitorHttp;
  WakeOnLanCfg? wolCfg;
  BmcCfg? bmc;          // ← this
}
```

## The account is a record, not two fields

BMCs are provisioned a rack at a time and answer to one directory or one factory password, so the normal case is many servers to one account. As columns on `server` that is the password typed once per machine and, on a rotation, changed once per machine — with no way to tell whether one was missed except by a machine that stops answering.

`bmc_credential` is its own table and its own sync root, keyed by a generated id with a `UNIQUE` name, so a rename is an `UPDATE` rather than something that detaches every server pointing at it. `server.bmc_cred_id` is `ON DELETE SET NULL`: losing an account must not lose the machines it opened, it leaves them with an address and nothing to log in with.

What stays on `BmcCfg` is what belongs to one device — the address, and the certificate fingerprint. The second one is the less obvious half: two BMCs never present the same certificate, so a fingerprint on a shared record would be the first device's used to verify the second, which is the check not happening at all.

A shared *account* is not a shared *BMC*. Several guests of one physical host would point at one device, where a power action on any of them cuts all of them and the reported `PowerState` is the host's rather than the guest's. That is a host/guest relation, not a storage question, and it is not modelled here.

## Trust on first use, not an ignore-the-certificate switch

Two places in this app already meet a self-signed HTTPS endpoint — a monitor agent and PVE — and both offer a switch that accepts anything. A BMC sits on a management network and holds power control, so it gets the `HostKeyVerifier` shape instead: the certificate is fetched and shown in the editor, and only the reviewed one is accepted afterwards.

The two halves are separate because the platform forces them to be. `badCertificateCallback` and dio's `validateCertificate` are both `bool Function(...)`, so nothing there can wait for an answer — the review is a step someone takes, and enforcement then answers by itself. Trusting whatever appears on the first *request* would be trust-on-first-use where the use already carries a password.

PVE and the monitor agent are left alone; changing published behaviour is its own decision.

## Vendor differences, and why they are discovered rather than assumed

Every one of these is something a client has to ask about. Hardcode any of them and it works on the machine it was written against.

| Finding | Consequence for the code |
| --- | --- |
| System ids differ: Supermicro `1`, Dell `System.Embedded.1`, OpenBMC `system` | Walk `Systems`/`Chassis`; never build a path |
| Redfish 2020.4 deprecated `Thermal`/`Power` for `ThermalSubsystem`/`Sensors`; Supermicro switched at X14, so X11–X13 are still old and transitional firmware carries **both** | Branch on which links `Chassis` has; prefer the new model |
| `ResetType@Redfish.AllowableValues` is advertised, which is not the same as implemented | Negotiate an intent against the list; offer nothing that maps to nothing |
| HPE iLO's `GracefulShutdown`/`GracefulRestart` are OS-dependent and iLO does not distinguish them | A `204` is acceptance, not a result — poll `PowerState` |
| Redfish sessions leak and BMCs allow few; enough leaked ones lock the interface out until timeout or a physical reset | One login per client, `DELETE` on the failure path as much as the normal one |
| Supermicro licenses gate firmware update and virtual media, not reads | A `401`/`403` on a sub-resource is an answer to report, not grounds to fail the fetch |
| A blade enclosure publishes one system per node | Only the first is shown, and `hasMultipleSystems` is what makes saying so possible |

## Layers

```
BmcCfg + BmcCredential    what the user configured
  ↓
RedfishClient             TLS trust, session lifetime, GET/POST
  ↓
RedfishDiscovery          one-time walk: which ids, which sensor model
  ↓
redfish/*.dart (pure)     JSON → models, no IO
  ↓
BmcNotifier (riverpod)    state, and a poll cycle of its own
```

Only `RedfishClient` touches the network, which is what lets every vendor difference above be tested against saved responses rather than hardware. The BMC is polled once a minute rather than on the status cycle: it is slow, it is answering about hardware, and riding the status poll would spend most of the device's capacity on numbers nobody is watching change.

## UI

- **Server detail** — a BMC card: power state, temperatures and fans, and what the service said it allows.
- **Power** — folded into the existing `power` button rather than a new one. The row holds eight entries and two of them should not be the same concept.
- **Server editor** — address, an account picker, and the certificate review.
- **Settings → BMC accounts** — the accounts, how many servers each opens, and the only place an account can be deleted.

## Notes for review

- `SchemaVersion` 6 → 7. `BmcColumnsMigration` guards each statement on `PRAGMA table_info`, and adds `bmc_cred_id` *with* its `REFERENCES` clause — without that a migrated install differs from a fresh one in the one way that is silent, the rule not firing.
- `BackupV2` carries `bmcCredentials`, merged before servers. A restored account reconciled onto an existing local id has every server's `bmc.credId` remapped to follow it, the same treatment `keyId` already had.
- The migration list moved to `migrations/all.dart`; `main.dart` and `hive_release_migration_test.dart` each had a copy.
- `BmcCfg` is deliberately **not** in `lib/hive/`'s `@GenerateAdapters`. No released build wrote one into a Hive box, so `HiveImport` has nothing to read, and a model in that list is a model whose shape can no longer change freely.
- Tests: 1337 pass. `test/bmc_redfish_e2e_test.dart` is opt-in and read-only — set `SBM_E2E_BMC_URL`/`_USER`/`_PWD`; it never resets a machine, and its header lists what it therefore cannot cover.

## IPMI

Out of scope here, and worth saying why rather than leaving it implied.

Coverage runs the other way from what it looks like: almost every BMC that speaks Redfish also speaks IPMI, but not the reverse — Supermicro X9, iDRAC7/8 below 2.30.30.30 and iLO4 have no usable Redfish at all, and X10/iDRAC8 have an early, partial one. That is a real slice of the hardware this feature is for. It is a second protocol behind the same `BmcCfg`, not a change to any of the above, and the Rust side (`ipmi-rs`) is where it would live.

Nothing about it is started, and keeping it on this PR's list would have quietly shaped the modelling here around it. `TODOS.md` records what it would take — including the part that is not a binding: an IPMI session is mutable state at an FFI boundary whose stated rule is that it holds none.


lollipopkit:
@winnowl ignore

lollipopkit:
@coderabbitai ignore

lollipopkit:
Code added, and taken out of draft. This is now design doc **plus** the first of the five steps.

## `BmcCfg`

Nested on `Spi` beside `wolCfg` rather than flattened into `ServerCustom` where PVE's three settings live: flat cannot express "not configured", which is the same reason the SSH fields were extracted into `SshCredential`.

## The certificate, and why it looks like this

BMCs are self-signed. The two places in this app that already meet a self-signed HTTPS endpoint — PVE and a monitor agent — offer a switch that accepts anything. A management interface holding power control is the worst place to extend that to.

The shape of the alternative was **forced by the platform, not chosen**, and it is worth stating because it is not what the issue or the design doc first assumed:

SSH host keys are verified by an *async* callback, which is how `HostKeyVerifier` can put the question to the user from inside it. Both TLS hooks are synchronous —

```dart
// HttpClient
void set badCertificateCallback(bool Function(X509Certificate, String, int)?);
// dio's IOHttpClientAdapter
typedef ValidateCertificate = bool Function(X509Certificate?, String, int);
```

— so nothing there can wait for an answer. Prompting from the verification path is not possible.

So it splits in two, and the split is better than a prompt would have been:

- **`fetchServerCert`** is the review step, driven from the edit page. It opens a socket for no purpose but to read what the far end presents, then destroys it. Nothing is sent, which is what makes accepting any certificate *for that one step* safe — no password reaches whatever answered.
- **`PinnedCert.accepts`** is enforcement, and answers alone. It refuses when nothing has been reviewed, deliberately: trust-on-first-*use* would mean trusting whatever turned up on the first request, and the first request already carries a password.

Enforcement will go on `badCertificateCallback` (during the handshake, before any bytes) rather than `validateCertificate` — the latter runs *after* the response arrives, by which point credentials have been sent. `validateCertificate` is still worth setting as a second check, for the case where a chain validates against a CA but is not the pinned certificate.

Written against no particular caller, so PVE and the monitor agent could adopt it later without moving their configuration. Not migrating them here — changing published behaviour is its own decision.

## Details worth flagging

- **SHA-256 of the DER form**, which needs `crypto` promoted from a transitive dependency. `dart:io` offers only `sha1` ready made, and that has no business carrying a trust decision.
- **An expired certificate is reported, not refused.** BMCs ship certificates that expired years ago; refusing those would refuse most of the hardware this exists for.
- The fingerprint is shown colon-separated and uppercase, which is how a BMC's own web UI prints it — the point is that someone can compare the two.

## Verification

`test/cert_pin_test.dart`, 16 cases, on the half that decides: matching, mismatching, nothing pinned, a null certificate, case, and a truncated pin that must not match by prefix. Plus `BmcCfg` address parsing, port defaulting per scheme, and that `copyWith` can *clear* a pin.

`flutter test` 912 pass. `flutter analyze lib test integration_test` clean.

Next: the Redfish client — service root, collection walking, session lifetime.

lollipopkit:
Second step pushed: the Redfish resources and discovery, still pure — no transport yet, which is deliberate. This is where every vendor difference lands, and none of it needs a BMC to test.

## Recorded shapes, three vendors

`test/redfish_test.dart` carries the service root, collections, system and chassis as **Dell iDRAC 9**, **Supermicro X11** and **OpenBMC** actually present them, and drives discovery against each. That is what makes the following testable today rather than after the hardware arrives.

## What the tests pin down

**Collections are walked, never built.** `Systems/1`, `Systems/System.Embedded.1` and `Systems/system` are all correct and none derives from the others — one test per vendor.

**The reset target comes from the action, not the system's path.** They agree on everything seen, but only one of them is what the service said. There is a test with the target pointing somewhere else entirely, to make sure the other is never used.

**`ResetType@Redfish.AllowableValues` decides, and can decide nothing.**

| service | restart resolves to |
| --- | --- |
| Dell iDRAC 9 (no `GracefulRestart`) | `ForceRestart` |
| Supermicro X11 (has it) | `GracefulRestart` |
| a service allowing only `On` | *nothing* — the intent is not offered |

An action with no `AllowableValues` at all is unusable rather than a licence to guess from the spec's enum: sending a type the service never claimed to take is how `Nmi` and `PowerCycle` bite.

**Both sensor models, and the new one wins.** That is what the deprecation means, and transitional firmware is exactly where both appear. A `ThermalSubsystem` with no `Sensors` collection counts as neither — the readings live in `Sensors` under the new model, so the link alone is a path to nothing.

**A refused chassis costs the sensors and nothing else.** Licensing gates parts of some services, so a 403 on one sub-resource is an answer about that resource, not grounds to call the service unusable. The test asserts the power half still comes back.

**A body that parses but has no `Systems` and no `Chassis` is not a service** — which is what a static host answering every path with its index page looks like. Same failure the monitor panel had against Cloudflare Pages.

## The line around resets

`ResetRequest` is **built and returned, never sent**. Which type, against which path, with which body is asserted; performing it is left to a person. That is the line `script_compat.rs` already draws around shutdown and reboot, and it is what keeps the suite from being able to power-cycle your machine.

`flutter test` 934 pass, `flutter analyze lib test integration_test` clean.

Next: the dio-backed transport — pinned certificate on `badCertificateCallback`, one login per client, `DELETE` of the session on every exit path.

lollipopkit:
Third step: the transport. The only part that touches the network — and it turned out to be testable without hardware after all.

## Tested against a real HTTPS server

`test/redfish_client_test.dart` starts an actual `HttpServer.bindSecure` on loopback with a self-signed certificate and speaks enough Redfish to answer. Pinning and session lifetime cannot be checked by reading a JSON map, and both are expensive to get wrong in ways that surface late: an unpinned certificate is a password handed to whoever answered, and a leaked session is a BMC that eventually refuses everyone.

The keypair is a throwaway `localhost` one in `test/fixtures/`, committed so the suite needs no `openssl` at run time.

## The certificate, enforced in two places

They run at different moments and only one is early enough:

| hook | when | what it is for |
| --- | --- | --- |
| `badCertificateCallback` | during the handshake, before a byte of the request exists | the one that matters — a password must not reach a certificate nobody vouched for |
| dio's `validateCertificate` | after the response arrives | too late to protect a credential; set for what the first never sees, a chain that validates against a real CA but is not the pinned certificate |

The tests assert the server saw **no login at all** when the pin was missing or wrong — not merely that the call failed. Failing after sending the password would also "fail".

## Sessions

One login per client, shared by whatever asks concurrently, deleted on close.

Removing the shared-login memo makes three concurrent reads create three sessions, which the test catches:

```
Expected: <1>
  Actual: <3>
```

Three is how a device that allows four ends up refusing everyone. Also covered: a probe of the service root uses no session at all (so asking "is this a BMC" leaves nothing behind), `close` is safe twice, and a client that never logged in still closes.

## Answers that mean something

`401` and `403` are read rather than thrown on. Licensing gates parts of some services, so a `403` names the resource it was about and the caller decides what that costs — the discovery test has an empty chassis collection and asserts the system still comes back.

## Where this leaves #1269

| | | offline verification |
| --- | --- | --- |
| 1 | `BmcCfg` + certificate pinning + edit page | done |
| 2 | Redfish resources + discovery | done, three vendor shapes |
| 3 | transport: TLS + sessions | done, against a real TLS server |
| 4 | the `bmc` card + slow poll cycle | next |
| 5 | power control | after that; the request is asserted, sending it stays manual |

`flutter test` 946 pass, `flutter analyze lib test integration_test` clean.

Still unverified and marked as such: anything about how a *real* BMC behaves. The fake here answers the way the specification says; the vendor differences that matter are the ones in step 2, which is why those are driven by recorded shapes rather than by this server.

lollipopkit:
Phase 1 is complete — all five steps. Retitled accordingly.

## Power control

An intent (restart, shut down, power on, force off, power cycle) is resolved against `ResetType@Redfish.AllowableValues`, and **an intent the service allows nothing for is not offered at all**. A button that fails when pressed is worse than one that was never there, and `Nmi`/`PowerCycle` are advertised-but-unimplemented often enough that this is not hypothetical.

One asymmetry worth stating: **a graceful shutdown has no fallback, on purpose.** `ForceOff` is not a shutdown, and substituting it would take a machine down hard when someone asked for the polite thing. Restart *does* fall back — `ForceRestart` is the same operation done less politely, and Dell offers no other.

The confirmation dialog names the **`ResetType` actually chosen**, not the intent, because they differ: a "restart" is `ForceRestart` on hardware with no graceful one, and that is worth seeing before agreeing.

**A `204` is not a result.** HPE documents that graceful operations depend on the OS and that iLO does not distinguish them, so `PowerState` is polled until it settles somewhere non-transitional. Accepted-but-unmoved is reported as exactly that — a toast saying "done" there would be reporting the request back as though it were the outcome.

## The line around resets

Nothing automated performs one. `test/bmc_power_test.dart` uses a transport that writes requests down instead of sending them, and its header says so in the same terms `ssh_e2e.rs` uses for shutdown and reboot. Every decision up to the send is asserted; the send is a person's.

## Phase 1, done

| | | verified by |
| --- | --- | --- |
| 1 | `BmcCfg` + certificate pinning + edit page | 16 cases, incl. a truncated pin that must not match by prefix |
| 2 | Redfish resources + discovery | Dell / Supermicro / OpenBMC recorded shapes, 22 cases |
| 3 | transport: TLS + sessions | a real self-signed HTTPS server, 12 cases |
| 4 | sensors, the card, its own poll cycle | both sensor models, 15 cases |
| 5 | power control | 11 cases, none of which send anything |

`flutter test` 972 pass, `flutter analyze lib test integration_test` clean, docs site builds.

## What is not verified

Both doc pages moved from "not shipped yet" to **"untested against real hardware"**, which is what is now true. Everything is handled against recorded responses and a local TLS server — enough to be sure of the *decisions*, not enough to be sure of a *machine*.

When you have the `.env` lines ready I will add the opt-in e2e: read-only, and it will read `ResetType@Redfish.AllowableValues` without ever posting to the action. Knowing the vendor and firmware would also tell me which sensor path your hardware actually takes, which is the one branch no fixture can settle.

lollipopkit:
Superseded by #1336. This one was marked merged against `58156d82`, the commit that merged `main` *into* `docs/bmc-redfish` — nothing here reached `main`, and the branch has carried on since.

PR #1329, opened 2026-08-21
feat(ios): several Linux systems at once, and the UI around them

lollipopkit:
Draft. Two months of the iOS Linux environment and the UI around it, on one branch: 35 commits, 101 files.

## Linux systems on the device

**Several at once, each in its own root.** The machine's root is a container of trees, one subdirectory per system, and nothing runs at that level — a session is rooted at its own subtree, the way a container is. Two Alpines can sit beside each other and both run: the profile id is a generated directory name, not the distribution, which is the case keying by distribution made impossible.

A profile is a directory plus the marker in it, and nothing else — no table, no setting listing what exists. So one deleted from disk cannot linger in a list, and a list cannot promise a tree that is not there.

**Selecting is not switching.** Nothing is deleted when the selection moves, sessions already running stay where they are, and a terminal records which system it is in so a restored tab reopens the one it was.

**Settings under Terminal > Linux**, named for Linux and not a distribution: the systems with rename/update/delete, the shell an interactive terminal starts, the mirror per distribution, and the resolvers written into the guest. The mirror and the resolvers had no setting before, so a device on a network that cannot reach `dl-cdn.alpinelinux.org` ended up with a Linux it could not install packages into and nothing that said so. The digest stays pinned in code while the mirror is a setting: a mirror decides where the bytes come from and never which bytes are accepted.

**A `chsh` for systems that have none.** Alpine ships none, and installing the real one would not help: it edits `/etc/passwd`, and nothing in this guest reads that. The stand-in writes the file that does decide — the same file the settings page reads — so there is one answer rather than two stores and a rule about which wins.

## The iOS engine

Needs lollipopkit/ShellBox#3 (merged): an absolute symlink target restarted from the machine root, so Alpine's `/bin/sh -> /bin/busybox` resolved to a `/bin` a rooted task cannot see, and `getcwd` reported the container path for what was the task's own `/`. Without the first, nothing in a profile could exec at all.

Around it: init no longer fork/execs a shell as fast as it can, the guest console is carried as bytes rather than through a `String` (anything outside ASCII did not survive the round trip), and `scripts/ensure-ish-libs.sh` fetches the static libraries built for the fork's exact revision — or builds them — from the Runner's build phases, so a checkout does not link yesterday's engine.

## Android

proot runs with `--link2symlink`, without which every package carrying hard links fails on exactly those entries. `wakelock_plus` is compiled at the Kotlin 2.1 language level, since its generated pigeon has no package declaration and Kotlin 2.2 rejects the imports that follow from that.

## Terminal input

One Cmd+V pastes once. Two paths were both pasting — a global `HardwareKeyboard` handler and the view's own shortcut — and the global one ran once per open terminal, so a key also reached terminals nobody was looking at. Paste is bracketed now, so an editor stops auto-indenting what arrives and a shell stops running the newlines.

## UI

A session line and a walkthrough for the virtual keys; the Agent header names the conversation and opens the list; settings content starts at the top with the tabs floating over it; card grids tightened; a page pushed in a tab animates across the status bar; the Agent tab and its history rail follow the theme.

## Verified on a device

iPad Pro, two Alpine trees under one machine root:

| | |
| --- | --- |
| `pwd` | `/` |
| `cd /tmp; pwd` | `/tmp` |
| `cd /..; pwd` | `/` — the root is a floor |
| `ls /` | the tree's own, with no sight of the container |
| second tree's `pwd` | `/`, not `-2` |
| `ls -l /bin/sh` | `-> /bin/busybox`, and it execs |
| `chsh -s /bin/ash` | takes effect in the next terminal, in that system only |
| `apk update` | reaches the mirror |

`flutter test` 1126 pass; `flutter analyze lib test integration_test` is clean apart from one unused import that predates this branch.

## Not done

- **The update button never appears on iOS.** `Rootfs.isOutdated` is still `isAndroid && …`, from when iOS recorded no version. It records one now, so this is a three-line change — held back because offering an update means downloading the release again and losing everything installed in that tree, which is a product decision.
- **No migration.** The container moved from `Application Support/alpine` to `.../linux`; an existing install reads as absent and is reinstalled. Deliberate, since none of this has shipped.
- **Isolation is one kernel's worth.** Shared PID space, shared network, one pty numbering, and the systems see each other through `/proc`. Stated in `sbm_ish.h` rather than implied. Real isolation would need a second process, which an App Store app cannot have — see TODOS.md.
- **Held keys do not repeat on iOS**, a platform gap rather than anything here: a backspace held 4.5s produces one down and one up. Recorded in TODOS.md with the measurement.

## Two things for the reviewer

- **`f7bcf366` and `3eb42ea0` do not build.** They delete `clipboard_chord.dart` without the import removal that went with it, which landed later in `3589f7d9`. The branch tip is fine; `git bisect` is not. Worth squashing on merge.
- **`packages/fl_lib` and `packages/flutter_pty` gitlinks move here.** Both need their own side merged first, or this points at commits nobody else can fetch.

## Summary by CodeRabbit

* **New Features**
  * Added support for managing multiple Linux profiles, distributions, shells, mirrors, and DNS settings.
  * Added a guided virtual-key walkthrough with localized help.
  * Improved SSH tabs, backup navigation, snippets, settings, file browsing, and empty states.
  * Enhanced iOS Live Activities with terminal counts and session details.

* **Bug Fixes**
  * Improved terminal text encoding, clipboard handling, safe-area layout, and command output limits.
  * Improved iOS Linux engine setup and profile lifecycle reliability.
  * Improved root filesystem installation and test-run consistency.

* **Documentation**
  * Added guidance for migration checksums, iOS builds, debugging, and Linux rebuilds.


lollipopkit:
@coderabbitai review

lollipopkit:
@coderabbitai review

PR #1332, opened 2026-08-22
feat(linux): Ubuntu and Rocky, several releases each, described by a signed manifest

lollipopkit:
The Linux-on-device feature grows from "an Alpine" to "a system you choose",
and stops carrying its pins compiled in.

## The feature is beta, and says so

A badge on the settings page and one dialog before the first install, with a
"don't show again" that is written only when the answer is yes — a box ticked
on a dialog that was then cancelled has agreed to nothing.

## Ubuntu and Rocky Linux

Both took engine work before they would install anything, all four bugs the
same shape — two distinct things conflated at some layer:

- `utimensat` dropped `AT_SYMLINK_NOFOLLOW`, so rpm could not stamp the
  `/usr/lib/.build-id/**` symlinks it lays down (ShellBox #20);
- `pselect6` read its optional sigmask without checking for null; glibc's
  `select()` passes null and musl does not, so every glibc program had a
  `select()` that could only answer EFAULT — apt read that as its download
  method having died (#21);
- `fchownat` had the same symlink-vs-target confusion as `utimensat` (#24);
- `TCGETS2` was unimplemented, which glibc uses to ask "is this a tty" where
  musl uses `TCGETS`, so Ubuntu's shell came up with no prompt (#26).

And one on this side: the tar reader conflated hardlinks with symlinks, so
115 of Ubuntu's coreutils applets unpacked as links pointing nowhere and `ls`
was not found.

## A signed manifest instead of compiled-in pins

Rocky republishes its container base every few weeks, so a pin shipped in a
binary names a file that is no longer there, and fixing it meant a release.
`lollipopkit/shellbox-rootfs` now publishes the images and a manifest
describing them; the app fetches it, verifies an Ed25519 signature against a
key compiled in, refuses a serial below the highest already accepted and
anything expired, and falls back to a re-verified cache and then to a bundled
copy. A mirror still decides only where the bytes come from.

That repository also normalises what it publishes — flattening Rocky's OCI
image so the app no longer walks one in Dart on a phone, and giving back
directory-write and owner-read bits that make a package manager fail partway
through unpacking in a guest whose root is a fiction.

## Several releases of one distribution

A distribution carries a list; the picker asks in two steps, system then
version. `branch` — apt's suite, apk's branch, dnf's major version — is what
the package manager reads and also what decides whether a later release is an
update: 24.04.3 and 24.04.4 are both `noble` and one replaces the other, while
26.04 is `resolute`. An install replaces the tree and destroys everything in
it, so an update never crosses that line. The series is recorded as a fourth
line of the marker; a three-line marker reads as "predates the question" and
falls back to the preferred release.

Offered now: Alpine 3.22.5 / 3.21.7, Ubuntu 26.04 / 24.04.4, Rocky 9.8 / 10.2.

## Deleting a system

Deleting used to need an app restart. A hung-up session's controlling terminal
was released only when the task was reaped, and an embedder does not reap — so
the pty stayed allocated, `/dev/pts` could not be unmounted, and the tree could
not be detached. `close_session` now disowns the terminal itself. The retry
loop that was waiting for something that would never happen is gone, the page
says what it is doing while it waits, and a system with a terminal open says
which one to close rather than failing after the fact.

## Testing

`flutter analyze lib test integration_test` clean; `flutter test` 1213 passed,
1 skipped. The engine fixes each carry a unit test in ShellBox, every one
checked by backing the fix out and watching it go red.

The two manifest fixtures are real: `ci_produced.json` is the output of
`shellbox-rootfs`'s own build, and `signed.json` is signed by the key the app
carries — a mocked signature would prove only that the code calls a verifier.

## Summary by CodeRabbit

* **New Features**
  * Added support for Alpine, Ubuntu, and Rocky Linux releases.
  * Added release-aware installation, updates, mirrors, package managers, and download details.
  * Added support for OCI images, compressed archives, hard links, and layered filesystems.
  * Added signed manifest refreshes with offline caching and validation.
* **Bug Fixes**
  * Prevented removal of Linux systems while terminals remain open.
  * Improved detachment behavior and terminal cleanup.
* **UI & Localization**
  * Added Linux beta labeling, warnings, and release selection.
  * Updated root filesystem messaging across supported languages.

### Changes

- **Signed rootfs manifest lifecycle and persisted trust state**: Introduces the schema-2 manifest model, bundled and cached manifest selection, signed network refresh, replay/expiry protection, bounded streaming fetches, and the associated setting-store persistence.
- **Distribution metadata, release selection, and network\-seeding contract**: Moves distro/release URLs, digests, layouts, compression, mirrors, package-manager metadata, and release-series behavior from code constants into the manifest while retaining typed platform-specific repository generation and profile marker compatibility.
- **Rootfs download, OCI extraction, archive fidelity, and filesystem safety**: Adds shared OCI layout/layer parsing and whiteout handling, extends Android and iOS extraction paths for multiple release formats, and preserves modes, links, cleanup, and guest initialization during installs.
- **iOS embedded ish engine ABI, lifecycle, sessions, and filesystem mounts**: Updates the native iOS shim and git-linked ish engine integration for optional availability, guest crash handling, real filesystem/device setup, profile attach/detach, pty sessions, byte I/O, and exit status behavior.
- **Linux profile management and user\-facing installation/removal flows**: Connects dynamic distro/release metadata and profile lifecycle to settings, terminal add pages, sidebar/profile UI, confirmation flows, updates, deletion, shell selection, and network configuration.
- **Localization assets and generated localization API**: Updates all supported ARB catalogs and generated localization Dart files for Linux beta/install/update/distribution-related UI text and keeps locale-specific accessors synchronized.
- **Dependency, repository guidance, and release/build configuration**: Adds the archive and Ed25519 dependencies, bundles the rootfs manifest asset, and updates repository guidance and build metadata for the new rootfs/OCI/trust functionality.


lollipopkit:
@coderabbitai review

PR #1334, opened 2026-08-22
fix(monitor): collect on NixOS, a NixOS module, and an onboarding skill

lollipopkit:
Two unrelated commits that happened to land together; the second is docs only.

## `fix(monitor)` — the agent collected nothing on NixOS

Found by installing it, not by reading paths. Built an aarch64 static musl
binary and installed it on NixOS 25.11: the service came up, the API answered
healthy, the database migrated — and **every collection cycle failed**.

```
ERROR ...monitoring::service: Failed to collect metrics:
  Status script error: No such file or directory (os error 2)
```

A systemd unit that sets no PATH gets systemd's compiled-in one, and that is a
per-distribution value. On most distributions it is the FHS list and this
change would be a no-op. NixOS builds systemd with its own store path, so the
service saw exactly one directory:

```
PATH=/nix/store/…-systemd-258.7/bin/
```

`sh` was not on PATH at all. It surfaces as a running agent with an empty
panel, which is the worst shape a failure can take.

Worth noting that `/bin/sh` *does* exist on NixOS — spelling the interpreter
absolutely would have started the script and then failed inside it on the
first `cat`. PATH is the level this belongs at, because the agent shells out
to whatever the command manifest names.

Measured after the fix: a clean install goes from a failure every cycle to
none, `status.sh` is written, and an alert rule fires on real data
(`cpu 78.08% (threshold: >=77%)`). Debian (systemd) and Alpine (OpenRC) were
reinstalled from the same package to check nothing moved — both still install
and collect.

### `--system`, and `monitor/nix/`

`--system` cannot work on NixOS at all: `/etc/systemd/system` is a read-only
symlink into the store, for root as much as anyone. It now says so *before*
trying, rather than surfacing a bare `Read-only file system` from the unit
write, and points at the module added here — a system service on this
distribution is `configuration.nix`'s job and there is no imperative
equivalent to offer.

The module carries a state directory and links the panel into it, because the
agent resolves `config.toml`, the SQLite file and `frontend/dist` relative to
its working directory and *writes* the first two. So that directory can be
neither the store nor a bare empty one. Migrations need nothing —
`sqlx::migrate!` embeds them at compile time.

Both `.nix` files are parse-checked and `npmDepsHash` is computed rather than
guessed. Neither has been through a real `nix build`; that is stated in the
commit message rather than implied away.

### Also

SSH discovery ran `/usr/bin/avahi-browse` by absolute path while calling `ip`
by name two lines above. An absolute `/usr/bin` is a guess about the
filesystem rather than about the tool, and it is wrong on NixOS, where only
`/bin/sh` and `/usr/bin/env` sit at an FHS path.

## `docs` — the onboarding skill

`.claude/skills/serverbox-onboarding/`: orientation for installing and using
the app, deploying the monitor agent, bootstrapping the Flutter + Rust + Node
environment, and the handful of facts that explain most "why can't I do X"
questions — chiefly that a server is reached over SSH **or** a monitor agent,
never both.

## Checks

- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean
- `monitor/frontend`: `eslint --max-warnings 0` clean, `svelte-check` 0 errors 0 warnings
- `flutter analyze lib test integration_test` — clean
- `website` build, `docs` locale parity — pass

Draft: the monitor half is verified on three distributions, but no
`monitor-v*` release exists yet, so the path users actually take
(`install.sh install`) is still untested end to end. See TODOS.md.

## Summary by CodeRabbit

* **New Features**
  * Added NixOS support for installing and running the ServerBox Monitor as a managed system service.
  * Added configurable service options, firewall handling, packaging, and frontend asset integration.
  * Added environment diagnostics to identify missing tools, dependencies, and setup issues.

* **Bug Fixes**
  * Improved Linux SSH discovery across systems where `avahi-browse` is installed outside standard paths.
  * Improved command discovery in generated Monitor services.
  * Added clearer handling for unsupported system-mode installation on NixOS.

* **Documentation**
  * Added comprehensive onboarding, installation, development, deployment, troubleshooting, and architecture guidance.
  * Documented current NixOS installation validation findings.

### DIFF
--- a/.claude/skills/serverbox-onboarding/SKILL.md
+++ b/.claude/skills/serverbox-onboarding/SKILL.md
@@ -1,0 +1,132 @@
+---
+name: serverbox-onboarding
+description: Orientation for ServerBox (flutter_server_box) — installing and using the app, deploying and configuring the ServerBox Monitor agent on a server, bootstrapping the Flutter + Rust + Node development environment, and explaining how the project actually works (SSH vs monitor HTTP, the shared Rust parser, SQLite storage, Riverpod state). Use this whenever someone asks how to install, set up, run, build, deploy, configure, contribute to, or understand any part of ServerBox or its monitor agent — including questions that never name the project outright, such as "how do I get this repo running", "pub get fails after cloning", "why is the terminal button missing on this server", "what does full_access actually grant", "where does the status data come from", or a first-time contributor asking where to start.
+---
+
+# ServerBox onboarding
+
+ServerBox is a Flutter app (iOS, Android, macOS, Linux, Windows, watchOS) for
+managing Linux, BSD and Windows servers: status charts, SSH terminal, SFTP,
+containers, processes, systemd. The repository is a monorepo:
+
+| Path | What it is |
+|---|---|
+| `lib/` | The Flutter app |
+| `crates/sbm_parser` | Shared status parser — one source of truth for the command manifest and the parsing, used by the app over FFI and by the monitor |
+| `crates/sbm_ffi` | flutter_rust_bridge binding crate, built into the app by `hook/build.dart` |
+| `crates/sbm_native` | Native per-platform sampler, monitor only |
+| `monitor/` | ServerBox Monitor: a Rust agent installed on a server, plus its Svelte panel |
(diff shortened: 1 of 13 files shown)


lollipopkit:
@coderabbitai review

PR #1336, opened 2026-08-23
feat(bmc): out-of-band management over Redfish

lollipopkit:
Supersedes #1316, which GitHub marked merged against the commit that merged `main` *into* this branch. Nothing from it reached `main` — same branch, same work, a PR that is actually open.

Closes part of #1269. Redfish only; IPMI is not here and is not planned for this PR — see the end.

The PR started as documentation and has since grown the implementation. What follows describes what is on the branch now.

## What a BMC is, in this app's terms

A side channel on `Spi`, beside `wolCfg`. Not a third `ServerConnectCredential`: that axis answers "where does this server's *status* come from" and is dispatched per transport, while a BMC answers about the machine when the host is absent. The two are also not exclusive — the normal case is SSH *and* a BMC on one record, which the SSH-xor-monitor `CHECK` would have refused.

```dart
class Spi {
  SshCredential? ssh;
  MonitorHttpCredential? monitorHttp;
  WakeOnLanCfg? wolCfg;
  BmcCfg? bmc;          // ← this
}
```

## The account is a record, not two fields

BMCs are provisioned a rack at a time and answer to one directory or one factory password, so the normal case is many servers to one account. As columns on `server` that is the password typed once per machine and, on a rotation, changed once per machine — with no way to tell whether one was missed except by a machine that stops answering.

`bmc_credential` is its own table and its own sync root, keyed by a generated id with a `UNIQUE` name, so a rename is an `UPDATE` rather than something that detaches every server pointing at it. `server.bmc_cred_id` is `ON DELETE SET NULL`: losing an account must not lose the machines it opened, it leaves them with an address and nothing to log in with.

What stays on `BmcCfg` is what belongs to one device — the address, and the certificate fingerprint. The second one is the less obvious half: two BMCs never present the same certificate, so a fingerprint on a shared record would be the first device's used to verify the second, which is the check not happening at all.

A shared *account* is not a shared *BMC*. Several guests of one physical host would point at one device, where a power action on any of them cuts all of them and the reported `PowerState` is the host's rather than the guest's. That is a host/guest relation, not a storage question, and it is not modelled here.

## Trust on first use, not an ignore-the-certificate switch

Two places in this app already meet a self-signed HTTPS endpoint — a monitor agent and PVE — and both offer a switch that accepts anything. A BMC sits on a management network and holds power control, so it gets the `HostKeyVerifier` shape instead: the certificate is fetched and shown in the editor, and only the reviewed one is accepted afterwards.

The two halves are separate because the platform forces them to be. `badCertificateCallback` and dio's `validateCertificate` are both `bool Function(...)`, so nothing there can wait for an answer — the review is a step someone takes, and enforcement then answers by itself. Trusting whatever appears on the first *request* would be trust-on-first-use where the use already carries a password.

PVE and the monitor agent are left alone; changing published behaviour is its own decision.

## Vendor differences, and why they are discovered rather than assumed

Every one of these is something a client has to ask about. Hardcode any of them and it works on the machine it was written against.

| Finding | Consequence for the code |
| --- | --- |
| System ids differ: Supermicro `1`, Dell `System.Embedded.1`, OpenBMC `system` | Walk `Systems`/`Chassis`; never build a path |
| Redfish 2020.4 deprecated `Thermal`/`Power` for `ThermalSubsystem`/`Sensors`; Supermicro switched at X14, so X11–X13 are still old and transitional firmware carries **both** | Branch on which links `Chassis` has; prefer the new model |
| `ResetType@Redfish.AllowableValues` is advertised, which is not the same as implemented | Negotiate an intent against the list; offer nothing that maps to nothing |
| HPE iLO's `GracefulShutdown`/`GracefulRestart` are OS-dependent and iLO does not distinguish them | A `204` is acceptance, not a result — poll `PowerState` |
| Redfish sessions leak and BMCs allow few; enough leaked ones lock the interface out until timeout or a physical reset | One login per client, `DELETE` on the failure path as much as the normal one |
| Supermicro licenses gate firmware update and virtual media, not reads | A `401`/`403` on a sub-resource is an answer to report, not grounds to fail the fetch |
| A blade enclosure publishes one system per node | Only the first is shown, and `hasMultipleSystems` is what makes saying so possible |

## Layers, and where the line is now

Everything that speaks the protocol lives in **`packages/redfish`**, a submodule with its own repository at [lollipopkit/redfish](https://github.com/lollipopkit/redfish) — pure Dart, Apache-2.0, no Flutter. pub.dev has no Redfish package, so it starts from nothing rather than a fork.

```
BmcCfg + BmcCredential    what the user configured          this app
  ↓
RedfishClient             TLS trust, sessions, GET/POST      package:redfish
RedfishDiscovery          one-time walk: ids, sensor model   package:redfish
resources / sensors       JSON → models, no IO               package:redfish
  ↓
BmcNotifier (riverpod)    state, and a poll cycle of its own this app
```

The extraction fixed a layering inversion rather than just moving files: `RedfishClient` took `BmcCfg` and `BmcCredential`, so a protocol client knew how this application stores credentials. It takes a base URL, an account and a certificate fingerprint now, and reports through a callback instead of depending on `fl_lib`.

Only `RedfishClient` touches the network, which is what lets every vendor difference above be tested against saved responses rather than hardware. 1261 tests here, 84 in the package, and the opt-in hardware test is over there too — what it exercises is the library. The BMC is polled once a minute rather than on the status cycle: it is slow, it is answering about hardware, and riding the status poll would spend most of the device's capacity on numbers nobody is watching change.

## UI

- **Server detail** — a BMC card: power state, temperatures and fans, and what the service said it allows.
- **Power** — folded into the existing `power` button rather than a new one. The row holds eight entries and two of them should not be the same concept.
- **Server editor** — address, an account picker, and the certificate review.
- **Settings → BMC accounts** — the accounts, how many servers each opens, and the only place an account can be deleted.

## Notes for review

- `SchemaVersion` 6 → 7. `BmcColumnsMigration` guards each statement on `PRAGMA table_info`, and adds `bmc_cred_id` *with* its `REFERENCES` clause — without that a migrated install differs from a fresh one in the one way that is silent, the rule not firing.
- `BackupV2` carries `bmcCredentials`, merged before servers. A restored account reconciled onto an existing local id has every server's `bmc.credId` remapped to follow it, the same treatment `keyId` already had.
- The migration list moved to `migrations/all.dart`; `main.dart` and `hive_release_migration_test.dart` each had a copy.
- `BmcCfg` is deliberately **not** in `lib/hive/`'s `@GenerateAdapters`. No released build wrote one into a Hive box, so `HiveImport` has nothing to read, and a model in that list is a model whose shape can no longer change freely.
- Tests: 1261 here, 84 in `packages/redfish`. `packages/redfish/test/e2e_test.dart` is opt-in and read-only — set `SBM_E2E_BMC_URL`/`_USER`/`_PWD`; it never resets a machine, and its header lists what it therefore cannot cover.

## Also on this branch, and not about BMC

Kept together rather than split because they were found while verifying this branch's Linux artifact, and they touch nothing the BMC work touches — one workflow file and three submodule pointers.

- **`fl_build` fetches its AppImage tooling from upstream.** `appimagetool` and the runtime came from `cdn.lpkt.cn`, a mirror with no published provenance. The runtime is byte-identical to what the mirror served; `appimagetool` moves from an unlabelled `continuous` snapshot to 1.9.1, the nearest tagged release. The newer `runtime-x86_64` was measured against the one in use and kept out: it mounts in every case the current one does and no more, is 45% larger, and prints three errors before succeeding on a host with fuse2 but no `fusermount3`.
- **The AppImage carried the app one level down.** `Exec=ServerBox` in the packaged `.desktop` file named nothing that existed — the executable was at `bundle/ServerBox`, and only `AppRun` knew. Running an AppImage always goes through `AppRun`, so it worked; desktop integration, which reads `Exec`, was never given a correct path. The comment above the copy already said `bundle/*`, which is the intent — `Process.run` starts `cp` with no shell, so the glob would have been a literal path, and dropping it nested the bundle instead.
- **A manual run can pick which jobs build.** `workflow_dispatch` built all six every time, so checking one thing meant waiting on five runners that were not the question. `jobs: linux` now runs `Build linux` alone. An unknown name is refused rather than silently skipping everything, and the value reaches bash through `env:` rather than `${{ }}`.
- **xterm to Unicode 16.** Merged in from `chore/bump-xterm`: 1042 code points now measure at the width their release says, combining marks and ZWJ emoji render instead of being dropped, and a fully-coloured 240x70 screen costs 1.0 ms a frame rather than 134.
- **`flutter_pty` resolves again.** It asked for `hooks ^1.0.0` while its own `native_toolchain_c ^0.19.0` requires `^2.0.0`. Dev-dependencies of a path dependency take no part in the root's solve, so only standalone work on that package was blocked and this repo's lockfile does not move.

Verified end to end rather than from the source: two dispatched Linux runs, artifacts extracted and executed on Ubuntu 24.04 (FUSE mount, `Exec` target present, no leftover `bundle/`, 12 bundled `.so` with no missing dependencies), and the resulting AppImage confirmed working on real hardware.

## IPMI

Out of scope here, and worth saying why rather than leaving it implied.

Coverage runs the other way from what it looks like: almost every BMC that speaks Redfish also speaks IPMI, but not the reverse — Supermicro X9, iDRAC7/8 below 2.30.30.30 and iLO4 have no usable Redfish at all, and X10/iDRAC8 have an early, partial one. That is a real slice of the hardware this feature is for. It is a second protocol behind the same `BmcCfg`, not a change to any of the above, and the Rust side (`ipmi-rs`) is where it would live.

Nothing about it is started, and keeping it on this PR's list would have quietly shaped the modelling here around it. `TODOS.md` records what it would take — including the part that is not a binding: an IPMI session is mutable state at an FFI boundary whose stated rule is that it holds none.

## Summary by CodeRabbit

* **New Features**
  * Added Redfish BMC support for server status, hardware sensors, certificate verification, and supported power actions.
  * Added reusable BMC credential management, including creation, editing, sharing, and deletion.
  * Added BMC configuration to server editing and backups.
  * Added localized BMC messaging and English/Chinese documentation.
  * Improved macOS build and release automation.

* **Documentation**
  * Added BMC setup, capabilities, limitations, and validation guidance.

* **Bug Fixes**
  * Improved backup restoration and credential reference handling.
  * Updated Ask AI protocol labels and contributor links.


lollipopkit:
@coderabbitai review

PR #1345, opened 2026-08-24
feat(server): identify a server's distribution, and mark it in the lists

lollipopkit:
Identifies which distribution a server runs, and draws its mark beside the name
wherever servers are listed. 41 commits; the identification and the marks are
most of it, and a handful of unrelated fixes rode along.

## Identification

`/etc/os-release`'s `ID=` rather than substrings in the `PRETTY_NAME` prose.
The old approach was an ordered list of needles where a derivative had to be
listed before its base — the order *was* the correctness, and it was already
wrong once ("Fedora CoreOS 40" read as Fedora).

The `sys` command now returns `ID`, `ID_LIKE` and `PRETTY_NAME` together,
reading `/etc/os-release` then `/usr/lib/os-release` in the precedence that
file specifies. Three sources, asked in order of what each is worth: exact
`ID`, then the prose, then `ID_LIKE`. The prose matcher stays — a remote too
old for os-release, and a `monitor` agent predating the field, send only that.

`Dist` grew from 13 cases to 65. Case names are the `{DIST}` contract in a
custom logo URL, so they are added and never renamed.

Touches `crates/sbm_parser` (command manifest + parsers), `crates/sbm_native`,
and the monitor's `SystemMetrics`. Verified end to end over SSH against Debian
13, Alpine 3.23 (busybox grep) and NixOS 25.11 — the last writes `ID_LIKE=""`,
which must not become a base named `""`.

## Marks

A small mark beside each server, in the lists, the pickers, the rails and the
files tab. Off by default; turning it on shows `assets/distro/README.md` and
waits three seconds before the accept button can be pressed. Off means no
pixels — `distIcon()` answers null and callers omit the slot.

Where a mark comes from, in order: an address the user configures
(`serverMarkUrl`, expanding `{DIST}` and `{BRIGHT}`), then one of four shipped
files, then a fallback icon — the penguin for a Linux, a server outline for
anything not known to be one. All drawn in the row's colour; the large logo on
the detail page is untouched and stays in its own colours.

`{DIST}` name overrides live in a key-value editor, because no correct table
exists: the file names belong to whichever collection the user pointed at.

## Which logos ship, and why only four

Two questions, and only the first is easy. Showing a mark to say which system a
server runs is nominative use and needs nobody's permission. *Shipping the file*
needs a copyright licence on the artwork, and almost no project grants one — a
trademark policy permitting referential use says nothing about copying a file.

Shipped: **debian** (LGPL-3+ / CC BY-SA 3.0), **gentoo** (CC BY-SA 2.5),
**nixos** (CC BY 4.0), **alpine** (no copyright subsists — simple geometry).
The four CC attributions are registered with `LicenseRegistry`, so they appear
under Settings → About → License.

Not shipped, with the reason recorded per project in `assets/distro/README.md`:
Ubuntu, Fedora and Arch reserve the artwork; **Rocky** licenses it freely but
forbids altering the mark "in any way", which is what drawing it in one colour
is; Red Hat, Raspberry Pi and Kali reserve logo use outright. openSUSE is the
trap worth knowing — its artwork repo is CC BY-SA 3.0 at the root and
`logos/official/LICENSE` says "These logos are all rights reserved."

## Also in here

- The window no longer resizes twice at launch on macOS (`MainFlutterWindow`
  put back a hardcoded 400x777 written in 2023, dead since the window size was
  persisted).
- `fl_build.json` gone, its one key into `pubspec.yaml`; `make.dart` gone; the
  macOS build folded into the analysis workflow.
- Virtual-key rows are configurable; the connection count shows on the server
  tab; snippets group by tag.
- Two review passes, twelve defects fixed.

## Testing

Eleven new test files. Worth singling out: `dist_asset_render_test.dart`
compiles each shipped mark the way flutter_svg does. It caught `debian.svg`
drawing nothing — Debian publishes a 2001 Illustrator export whose
`xmlns:x="http://…"` reaches `double.parse` — which four other assertions
passed straight over, because flutter_svg fails silently.

`flutter analyze lib test integration_test` clean, `cargo test --workspace`
clean, `flutter test` 1455 passing (the one failure is the environment-gated
Windows SSH e2e).

Depends on fl_lib `afc9bf8`, already on its main.

## Summary by CodeRabbit

* **New Features**
  * Added distribution detection, icons, custom marks, logo URLs, name mappings, and licensing notices.
  * Added bulk server controls, connection badges, tab menus, snippet grouping, and close-all terminal actions.
  * Added configurable virtual-key rows with mobile paging.
  * Added grouped Ask AI and Agent shell settings.
  * Added Linux distribution identifiers and GPU vendor details to monitoring.

* **Bug Fixes**
  * Improved backup, settings, virtual-key, and known-host migration compatibility.
  * Improved logo URL validation, fallback behavior, and restoration reliability.
  * Preserved desktop window size and improved CPU, disk, and GPU metric handling.

* **Documentation**
  * Updated build guidance and distribution-mark licensing documentation.

### Changes

- **Local SQLite/Drift schema, migrations, and grouped settings**: Moves entity and preference persistence onto the expanded Drift schema, adds migrations v8–v14, grouped Agent/AI settings with field adapters, server distribution cache, and backup/restore hooks.
- **Distribution identity, marks, and user\-visible licensing**: Adds explicit os-release identity and fallback matching, persistent per-server observations, bundled/configurable marks, URL handling, licensing and consent UI.
- **Shared status protocol, native sampling, and monitor metrics**: Synchronizes command manifests and parsers, adds os-release fields and extended scheduling, moves core Linux sampling native, bounds subprocesses, and updates monitor API/rules/frontend contracts.
- **Server identity, transport state, scripts, and SSH virtual\-key compatibility**: Changes server model consumers and asynchronous status handling, extended polling, custom-command migration, virtual-key persistence, and SSH/home-tab restore behavior.
- **Application settings, navigation, AI/Agent consumers, and platform presentation**: Connects the new settings shapes to AI/Agent and settings UI, navigation/pane persistence, platform window/menu integration, and runtime metadata.
- **CI, release packaging, build hooks, and iOS embedded engine**: Consolidates Flutter setup and macOS CI, changes release selection/publication and artifact validation, removes obsolete build configuration, and builds/links the optional iOS Linux engine/rootfs.
- **Documentation, repository guidance, and monitor/development metadata**: Updates development instructions, architecture/building documentation, monitor guidance, asset attribution, and removes obsolete TODO metadata.


lollipopkit:
@coderabbitai review

PR #1346, opened 2026-08-25
feat: label screenshots with the app name, and cover the app switcher card

lollipopkit:
Two things an app can do about being looked at while it is not being used, plus three fixes that were in the way.

## Screenshot label

The Dynamic Island is composited by the system above every app window, but a screenshot captures the app's own layers and not that overlay. The strip of screen behind the pill is therefore invisible in use and visible in every screenshot a user shares, which is where the app's name now goes, as a badge in the app's own theme colours. The colours are pushed from Dart on every theme change, since a native view cannot read a `ColorScheme`.

Portrait Dynamic Island devices only. In landscape the pill moves to the long edge, and a notch is a different shape, so on either the badge would sit in plain view instead of behind something.

The pill's top offset is **not** constant across models — 14 Pro through 16 sit at y≈11 under a 59pt safe-area inset, 16 Pro and 17 at y≈14 under 62pt, iPhone Air lower still — but the gap *below* it is. So the geometry anchors to the bottom of the safe area rather than to a hardcoded y or a table of device identifiers, either of which the next model would falsify.

**Known gap:** while the Dynamic Island is active (Live Activity, call, screen recording, mic/camera indicator) iOS renders the pill into the framebuffer for real and the badge is covered. This app has a terminal Live Activity, so a screenshot taken during an SSH session will not carry the name. Not controllable from the app.

## App switcher cover

Off by default; the switch is in Settings → App, under biometric auth.

| | iOS | Android |
|---|---|---|
| Mechanism | `UIVisualEffectView` over the Flutter window | `FLAG_SECURE` |
| Effect | Real blur, visible on screen | Blanks the recents thumbnail and blocks screenshots; the running UI is untouched |
| Applied | `willDeactivate` | `onPause` |

Android is not a visual cover, and deliberately so: Flutter draws into a `SurfaceView` that `RenderEffect` cannot reach, and any cover that has to render a frame races the system's recents capture, which the flag does not. The cost is that Android cannot hide anything from someone looking at the screen — only from the thumbnail and from screenshots.

Coming back, the cover is held until Dart has decided whether a biometric lock is coming. That closes a window which predates this branch: `didActivate` to `LocalAuthPage` being pushed used to show the real UI for several frames. It is released just *before* that push — on iOS the cover is a view over the whole Flutter window, so it is above every route drawn inside it, and holding it through the lock screen would hide the very thing it was being held for.

## Not yet verified

Both need a look on a real device; I could not confirm either from here.

- **The Dynamic Island badge has never been seen rendering.** On the simulator, oversizing it to 64pt tall (the pill is 37pt) still produced no pixels in a screenshot scan, so the overlay window is not drawing. The colours do reach the native side. My unconfirmed suspicion is that a non-key overlay window reports `safeAreaInsets.top` as 0, so the code decides the device has no Dynamic Island and hides the badge.
- **Whether the iOS blur survives into the switcher snapshot.** iOS may downgrade a `UIVisualEffectView` to flat colour when taking it. If it does, there is no way around it: Flutter renders to a Metal layer, so `drawHierarchy` captures nothing and the usual snapshot + `CIGaussianBlur` approach does not apply.

## Unrelated fixes

- `fix(android)` — `WRITE_EXTERNAL_STORAGE` declares `maxSdkVersion=32` here and 28 in `camera_android_camerax`; the merger refuses to pick, `:app:processDebugMainManifest` fails, and no APK builds at all. `tools:replace` keeps this app's own answer. **This was blocking any Android build on `main`** and can be cherry-picked out if it should land separately.
- `fix(settings)` — the floating tab bar's shadow was clipped above and below. Its horizontal `SingleChildScrollView` clips to a viewport exactly as tall as the bar, so the sides survived and the top and bottom stopped square. The room is now padding inside that viewport; the clip stays, because the horizontal axis still needs it.
- `fix(files)` — the empty-directory mark was a list item, so it sat at the top at icon height and read as the first entry of a directory that has none. It now centres in what is left under the `..` row.

## Testing

`flutter analyze` clean, `flutter test` 1507 passing. iOS device/simulator and Android APK all build. The one failing test, `windows_install_ssh_e2e_test.dart`, is a pre-existing opt-in SSH e2e that wants `SBM_E2E_SSH_KEY_PASSPHRASE`; it is unrelated and arguably should skip rather than fail when the key cannot be loaded.

## Summary by CodeRabbit

- **New Features**
  - Added optional privacy protection that blurs app content in the app switcher and protects it while inactive or locked.
  - Added theme-aware Dynamic Island branding on supported iPhones.
  - Added privacy protection controls to mobile settings.

- **Bug Fixes**
  - Improved file browser selection, deletion, navigation, naming, and empty-folder displays.
  - Improved path normalization and root-boundary handling.
  - Prevented stale watch updates and invalid server selections.
  - Refined settings tab positioning, spacing, and overflow behavior.

- **Localization**
  - Added privacy protection translations across supported languages.

### Changes

- **Cross\-platform privacy cover, Dynamic Island branding, and lifecycle coordination**: Adds native privacy-cover behavior on Android and iOS, theme-colored Dynamic Island branding, Dart method-channel/settings integration, biometric resume locking, and launch/lifecycle wiring.
- **Watch synchronization protocol, token lifecycle, and WatchConnectivity delivery**: Moves watch selection and legacy URL compatibility into settings, issues/reuses/revokes scoped monitor tokens, pushes durable and realtime payloads, adds revision ordering, and makes the watch app consume all delivery paths.
- **BrowsePath confinement and file\-browser navigation/mutation contract**: Separates browser root from current path, normalizes POSIX/Windows paths and dot segments, adds bounded back history, and routes file-browser navigation, pickers, listing persistence, and mutations through the new model.
- **Home tab identity, settings navigation tree, and responsive presentation**: Persists the last home tab by stable name with legacy positional fallback, publishes tab identity, restructures settings into hierarchical wide/narrow navigation with declarative pages, and adds responsive tab-bar presentation and guides.
- **Localization resources and generated localization API**: Adds or updates user-facing strings for privacy, watch/accessory settings, file browsing, settings navigation, and related features across ARB sources and generated locale implementations.
- **Redfish package pointer and repository build integration**: Includes the changed packages/redfish repository pointer as part of the reviewed change set.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -5,9 +5,14 @@
     <uses-permission
         android:name="android.permission.READ_EXTERNAL_STORAGE"
         android:maxSdkVersion="32" />
+    <!-- `tools:replace` because camera_android_camerax declares this same
+         permission with maxSdkVersion 28 and the merger refuses to pick. 32 is
+         this app's own answer: it writes downloads and SFTP transfers, which
+         scoped storage only takes over at 33. -->
     <uses-permission
         android:name="android.permission.WRITE_EXTERNAL_STORAGE"
-        android:maxSdkVersion="32" />
+        android:maxSdkVersion="32"
+        tools:replace="android:maxSdkVersion" />
     <uses-permission android:name="android.permission.USE_BIOMETRIC" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_DATA_SYNC" />
--- a/android/app/src/main/kotlin/tech/lolli/toolbox/MainActivity.kt
+++ b/android/app/src/main/kotlin/tech/lolli/toolbox/MainActivity.kt
@@ -4,6 +4,7 @@ import android.content.Intent
 import android.content.pm.PackageManager
 import android.os.Build
 import android.os.Bundle
+import android.view.WindowManager
 import android.Manifest
 import android.content.BroadcastReceiver
 import android.content.Context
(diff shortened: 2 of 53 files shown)


lollipopkit:
@coderabbitai review

PR #1348, opened 2026-08-25
feat: reach a server's files over scp, and keep Android alive in the background

### LINKED ISSUE #1288: 请添加新功能scp
嵌入式老设备设备只支持scp，请添加。

lollipopkit:
## What this changes

Two independent changes that happen to share a branch, plus a submodule bump.
They touch no common code and can be reviewed separately.

### `feat(file): reach a server's files over scp` — Fixes #1288

SFTP is a subsystem sshd has to be built and configured to offer, and the
machines this app is asked to reach do not always offer it — an OpenWrt router
running dropbear, an embedded box whose firmware ships one static `scp`. Those
hosts still run commands, and running a command is all SCP needs.

`ScpFileBackend` is the third `FileBackend`, beside local and SFTP. The split is
the protocol's own: SCP moves the bytes of one file and says nothing else, so
`read`/`write` go over `scp -f`/`scp -t` and everything else — list, stat,
mkdir, rm, mv, chmod — is a shell command. The listing command and its parser
move to `shell_file_ops.dart`, shared with the SFTP backend's sudo escalation so
the two cannot drift.

The wire protocol is written against a `ScpChannel` rather than an `SSHSession`,
so it is exercised by a scripted remote in tests rather than needing a server.

Which protocol a host speaks is `SshCredential.fileTransport`, stored per server
(schema **v15**) and chosen in the editor under **More**. Not probed: opening
SFTP and falling back would take a dropped link or a locked account for a host
with no subsystem, and answer the second failure instead of the first. The
browser's error view names the alternative when an SFTP session is what would
not open.

Two consequences of SCP worth knowing at review time: `size` is a hint
everywhere else in `FileBackend` and a **contract** here, because the sink is
told the length before the bytes start — a write with no size is spooled to
measure it; and `read`'s offset is honoured by discarding locally, since the
protocol cannot start anywhere but zero.

Two fixes that are not about SCP but are in the same commit:

- **A staged replace no longer changes a file's permissions.** Every `write`
  stages beside the destination and renames, and the staged copy carries the far
  side's umask onto it — so saving an edit to a 0755 script left it 0644, and
  replacing a 0600 file made it world-readable. The monitor agent already did
  this on its side; `carryModeToStaging` is now the one implementation for the
  SFTP and SCP backends and for the specialised SFTP upload, which is the path
  the editor's save-back takes.
- **`SshCredential` is frozen out of `@GenerateAdapters`.** Hive is the old
  database: no box has ever carried a field added after the last Hive release,
  so an adapter that knows about one describes a shape nothing is in. The
  generated one still compiled here only because the new field has a default.

### `fix(android): keep the process alive in the background` — see #1287

What keeps an Android process out of the freezer is a foreground service, and
this app could not start one. `MethodChans.startService`/`stopService` returned
at their first line unless a `fgService` setting was true — a switch that was
commented out of the Android settings page (`// _buildFgService(),`) and then
deleted with that page in 9ac86664. It was false for every install, so both
calls did nothing for over a year.

The only thing that ever promoted the service was `updateSessions`, as a side
effect. The moment the session list went empty that side effect ran the other
way: the native side cancelled its notification, the process left the
foreground, and Android 12+ refuses to let a backgrounded app start a foreground
service again. **One disconnect in the background was therefore permanent.**

`bgRun` is the one switch now and means what it has always said.
`TermSessionManager` owns the decision, under three rules: start on `inactive`
rather than `paused` (the last moment Android allows the request); never stop
while backgrounded (the one direction that cannot be undone from there, hence
`keepAlive` in the payload); and say so when there is no notification
permission, since without it none of it can happen.

The terminal's keep-alive had a second defect, which explains what the reporter
found for themselves — that the drops stopped once they deleted the servers that
could not connect. A resume ran `_checkConnectionHealth(immediate: true)`, which
meant "one missed reply is the verdict", while `home.dart` was on the same frame
starting a refresh of every configured server. An SSH handshake is pure Dart on
that isolate, so each unreachable server costs a connect timeout of its
attention, and a few were enough to push one ping past its ten seconds and close
a session that was fine. It now asks as many times as the timer would before
concluding anything. A suspected loss then reconnects rather than asking — that
was gated on having a tmux session to re-attach, but a link dropped while the
phone was in a pocket has no tmux, and "go back?" is not an answer to it.

## How it was tested

- `flutter analyze lib test integration_test` clean.
- `flutter test` green except `windows_install_ssh_e2e_test.dart`, which is
  gated on `SBM_E2E_SSH_*` and already failed before this branch.
- New tests: `scp_protocol_test.dart` (13 cases against a scripted remote —
  refusals, a host with no `scp`, a login banner ahead of the stream, a
  connection that dies mid-file, offset, over- and under-long streams, stderr
  beating a bare exit code), `carry_mode_test.dart`, `shell_file_ops_test.dart`,
  `m014_ssh_file_transport_test.dart`.
- `./gradlew :app:compileDebugKotlin` → BUILD SUCCESSFUL.

**Not verified on hardware.** The Android half is static verification only —
freezer behaviour, the `inactive` timing and the notification-permission flow
all need a device, and the reporter's ColorOS is the case that matters. I would
keep #1287 open until someone confirms it there, which is why this does not
auto-close it.

## Checklist

- [x] `make analyze` and `make test` pass
- [x] `make gen` was run, if any model / ARB file changed
- [ ] `cargo test --workspace` passes, if anything under `crates/` or `monitor/` changed — nothing there changed
- [x] No formatter was run over untouched code

## Summary by CodeRabbit

* **New Features**
  * Added SCP support with SFTP/SCP selection for SSH file transfers.
  * Added a remote file manager with uploads, downloads, folders, renaming, permissions, and deletion.
  * Added Android notification-permission guidance and improved background session handling.
  * Improved SSH reconnection, Linux profile management, and virtual-key availability.

* **Bug Fixes**
  * Improved file permissions, symbolic links, path validation, hidden-file search, and directory transfers.
  * Improved transfer reliability, service lifecycle, and server-edit error handling.
  * Improved legacy SSH credential compatibility and root filesystem handling.

* **Localization**
  * Added related guidance and file-management translations in multiple languages.

### Changes

- **SSH file transport abstraction and SCP protocol**: Introduces a transport-neutral FileBackend contract, SCP-over-SSH plus shell metadata operations, and broadens SFTP behavior to share atomic writes, permissions, symlink handling, escalation, and timeouts.
- **Cross\-backend copy and transfer lifecycle**: Generalizes transfers to local, monitor, SFTP, and SCP endpoints, adds directory/tree copying, staging notifications and cleanup, and changes isolate selection and cancellation behavior.
- **SSH credential transport persistence and server identity**: Adds per-server SFTP/SCP selection, key-reference/equality normalization, server-store persistence, generated serialization, editor controls, and migration of the server schema.
- **File browser, tabs, and storage UI contract**: Makes the browser operate on the backend abstraction, adds server/session restoration and cross-end send flows, exposes SCP fallback/escalation, and updates storage pages and transfer presentation.
- **Rootfs manifest, installation, and lifecycle safety**: Tightens rootfs manifest parsing/version comparison, adds signed/fetched manifest data flow context, and hardens Android/iOS profile scanning, installation, extraction, deletion, and platform abstraction.
- **Android foreground service, native bridge, and process control**: Changes Android foreground-service keep-alive/restart/notification behavior, native method channels and permission callbacks, privacy protection, and process-tree termination.
- **SSH/session and application integration consumers**: Updates server/session providers, AI/global agent tools, SSH pages, home/settings consumers, and generated provider artifacts to account for changed server/file/session contracts.
- **Localization, generated artifacts, dependencies, and repository configuration**: Adds transport/rootfs/storage/service-facing strings and updates generated localization/adapter/provider artifacts plus submodule/configuration metadata.


lollipopkit:
@coderabbitai review

lollipopkit:
@coderabbitai review

lollipopkit:
@coderabbitai review

PR #1349, opened 2026-08-26
feat: rebuild the watch and home widgets on the agent's authenticated API

### LINKED ISSUE #951: Android Widget not Connection
**Describe the bug**

The Android Widget just shows "Error Network error. Please check your Network connection" even though I've checked on multiple devices, the status page is showing as described in the ServerBoxMonitor wiki.

**To Reproduce**

1. Install ServerBoxMonitor
2. Add new Android Widget
3. Enter Server ip:Port/status

**Desired Results**

- A Widget showing the Servers stats

**Actual Results**

- Network Error

**Device**

- Galaxy S23 Ultra, Android 16, OneUi 8, App v1270 from Fdroid

lollipopkit:
Rebuilds the watch app and both home-screen widgets around the agent's
authenticated API, retires the endpoint they used to read, and lets one server
carry SSH and a monitor agent at the same time.

Closes #951.

## Why these are one change

A widget could never draw a trend, and the reason was the endpoint: `/status`
answered unauthenticated, with values preformatted as strings — `"1.3g / 1.9g"`,
`"31.7%"` — and no history at all. Half of `StatusWidget.swift` was machinery
for parsing those strings back into numbers so a gauge could be drawn from
them, and no amount of it could produce a series. Charts mean
`/api/v1/metrics` and `/metrics/history`, those need a credential, and a
credential needs somewhere to live on each platform. Everything else follows
from that.

## What changed

**The credential.** `POST /api/v1/watch-token` mints a read-only, revocable
token per `client_id` — `watch:<id>` and `widget:<id>` are deliberately
different, so revoking one surface leaves the other alone. It reaches exactly
three endpoints, and that limit is the agent's route table rather than
anything in the token. `tests/watch_token_scope.rs` now holds it, mounting the
real route table so a hand-built test router cannot pass for the shipped one.
Verified by injecting the mistake it exists to catch.

Also fixes a latent bug: the agent issues these for 90 days and answers with
`expires_at`, which the app threw away and then reused the token forever. On
the 91st day a paired watch started answering 401, on a device with nothing to
report it to. Tokens are now replaced inside a two-week window.

**Storage differs by what each platform allows.** The watch has its own
Keychain. The Android widget provider runs in the app's process, so it reads
app-private storage directly, with the token encrypted under an
AndroidKeyStore key. The iOS widget is a separate extension and needs a shared
Keychain group — placed *second* in `keychain-access-groups`, because an item
added without naming a group lands in the first entry, `flutter_secure_storage`
names none, and that is how the database encryption key is stored.

**Watch.** Servers page vertically under the Digital Crown; one server's
metrics page across. Each metric gets the whole screen instead of three charts
34pt tall stacked in a column. Disk gains a chart. The complication follows
both axes rather than being fixed to CPU. Sync is automatic — every server with
an agent, minus an exclusion list, since an opt-in list is somewhere to forget
a server and a forgotten server looks exactly like a broken watch.

**Home widgets.** Both pick a server from the list the app publishes. Small,
medium and large; readings as text, or one, two or four charts. iOS uses
AppIntents, which is why that extension alone moved to iOS 17 (the app stays
at 15). Android is one resizable provider deciding the layout at draw time
from the space the launcher actually gave it, with every chart panel in one
bitmap — four separate ones is four times a binder payload with a limit a
widget cannot recover from.

**SSH and monitor together.** They were exclusive, enforced three times over.
Both is a real configuration: an agent for status without holding a shell
open, sshd for what the agent has no endpoint for. `preferredTransport` orders
them, capabilities are the union of both, and `ensureExec` falls through on
failure. Schema v18 is a table rebuild, since SQLite cannot drop the CHECK
constraint that enforced the exclusivity.

## Two things worth a second look

**`usesCleartextTraffic` needed its justification rewritten.** It was added
last week on the grounds that the widget only ever fetched an unauthenticated
endpoint. It now carries a token, so that reasoning is gone. The flag is
process-wide permission; whether a given server's credential may go over an
unencrypted wire is `allowInsecure`, published per server and checked per
request.

**Manual URL entry in the Android widget is gone, and I did not keep it.** The
original ask was to keep a free-text field for addresses not configured in the
app. That made sense against `/status`, which needed no credential. Against
the authenticated API an address alone is useless — reaching one needs a
login — and a login typed into a widget configuration dialog is a second place
credentials live. Say if you want it back in some form.

## Upgrade cost, stated plainly

- iOS 15/16 users lose the home widget entirely: the extension will not load
  below 17. They were also going to lose it to the `/status` removal.
- Anyone who hand-typed a `/status` address loses that configuration. There is
  nothing to convert it into, so schema v17 detects the case and raises a
  one-time dialog explaining it.
- The agent answers 410 on `/status` for one release before the route goes, so
  a client too old to have another path gets an explanation rather than a 404.

## Verification

`flutter test` 1709 passing. `cargo test --workspace` green. Android
`assembleDebug`, iOS `StatusWidgetExtension`, watchOS `WatchApp` and
`WatchWidgetExtension` all build. `flutter analyze lib test` clean.

Two pre-existing failures are untouched and unrelated: `sbm_parser`'s
`ssh_e2e` and `windows_install_ssh_e2e_test.dart`, both opt-in tests enabled
by `.env` whose hosts are not currently reachable.

Not verified by me, and the reason this wants a real device: every widget and
complication layout, the iOS configuration sheet, the Keychain access group
against a real provisioning profile, and the watch's two-axis paging.

## Summary by CodeRabbit

* **New Features**
  * Configure servers with SSH, Monitor HTTP, or both, including preferred and fallback connections.
  * Set up authenticated watch and home-screen widgets using saved server credentials.
  * Added dedicated small and medium widgets with metric views, charts, refresh controls, and history.
  * Watch app now supports server paging and selectable CPU, memory, disk, network, and overview charts.
  * Added localized widget configuration, status, accessibility, and error messages.

* **Bug Fixes**
  * Improved connection fallback, credential renewal, secure storage, transport validation, widget sizing, and terminal focus handling.
  * Legacy unauthenticated status links now provide migration guidance.

### Changes

- **Server transport model, persistence, and editor behavior**: Allows a server to retain both SSH and monitor transports, persists a preferred transport, updates validation/serialization/store reconstruction, and exposes the new choices in the server editor and related generated/provider code.
- **Scoped monitor credentials and watch/widget synchronization contract**: Introduces expiring, endpoint-bound, independently scoped read-only tokens; synchronizes watch and home-widget server payloads; wires lifecycle, store-change, deletion, renewal, and revocation paths.
- **Android home\-screen widget implementation and platform bridge**: Replaces manually entered legacy status URLs with app-published server selection, encrypted Keystore-backed tokens, authenticated metrics/history requests, responsive chart rendering, and widget configuration lifecycle handling.
- **iOS home\-screen widget, shared containers, and target integration**: Migrates the iOS widget to AppIntent server selection and authenticated metrics/history, adds App Group/Keychain storage and native publication, and updates Xcode target metadata and entitlements.
- **Watch app, connectivity, and complication compatibility**: Updates the watch transport/client and app/widget presentation to consume the new authenticated server payload, track selected server/chart state, and render expanded metric/history data.
- **Storage migrations, legacy retirement, and user\-facing migration notice**: Adds schema steps for watch-selection inversion, legacy URL/widget preference retirement, and the dual-transport table rebuild, with permanent regression tests and notice/localization plumbing.
- **Build, deployment, repository hygiene, and cross\-cutting runtime changes**: Changes platform/build guidance and assorted generated/runtime/configuration artifacts that can affect packaging, startup ordering, distribution, and release contents.


lollipopkit:
@coderabbitai review

PR #1375, opened 2026-08-28
feat(ssh): compute SSH crypto natively, and stop reparsing private keys

lollipopkit:
Moves the SSH crypto that runs on the isolate drawing frames into `sbm_ffi`, and
stops parsing the same private key once per connection.

Depends on lollipopkit/dartssh2#22 and #23, both merged; the gitlink moves to
`ec28150`.

## Why, and how it was measured

**Benchmark in AOT, not in `flutter test`.** pointycastle is roughly 4x slower
compiled AOT than under the JIT — 27 MB/s against 108 for aes256-ctr alone — so
a figure taken in the test VM understates a release build, and understates it on
one side only. Everything below is `dart compile exe` on a pure-Dart harness
that cross-checks both implementations agree before timing either.

Record layer, aes256-ctr with hmac-sha2-256, 8 MiB per packet size:

| packet | pointycastle | native | |
| --- | --- | --- | --- |
| 64 B | 5.5 MB/s | 36 MB/s | 6.5x |
| 1 KiB | 13.5 MB/s | 261 MB/s | 19x |
| 32 KiB | 15 MB/s | 486 MB/s | 32x |

The 64 B row is the one that had to be checked: a crossing per packet was the
reason to doubt this pays off at keystroke sizes, and it wins there anyway.

Per connection, ms per operation:

| operation | Dart | native | |
| --- | --- | --- | --- |
| x25519 keypair | 0.329 | 0.008 | 40x |
| x25519 shared secret | 0.317 | 0.019 | 17x |
| ed25519 sign | 0.609 | 0.015 | 41x |
| ed25519 verify | 1.215 | 0.019 | 64x |
| ecdsa p-256 sign | 2.793 | 0.082 | 34x |
| ecdsa p-256 verify | 3.900 | 0.084 | 46x |
| bcrypt_pbkdf, 16 rounds | 173.6 | 82.4 | 2.1x |

The record cipher matters because it runs on every packet for as long as a
session lives; a terminal streaming output was capped at ~15 MB/s *while holding
the UI isolate*. The per-connection operations matter for a different reason: a
frame does not care about the total over a launch, it cares whether one
synchronous call fits inside it. A 120 Hz frame is 8.3 ms, so an ECDSA host key
check took half of one and two connections landing together dropped it.

x25519 also removes isolates rather than just time. `SSHKexX25519.createAsync`
and `computeSecretAsync` exist to keep scalar multiplication off this isolate by
running it in `Isolate.run` — two spawns per handshake, twenty when ten servers
connect at launch. At 8 µs there is nothing left to hide.

bcrypt_pbkdf is the one where the estimate was wrong: 2.1x, not the ~8x guessed
from the other ratios. Kept because it halves a wait the user sits in front of,
but it is not in the same class as the rest.

## What is deliberately not covered

- **RSA.** The pure-Rust `rsa` crate has no stable release and carries
  [RUSTSEC-2023-0071](https://rustsec.org/advisories/RUSTSEC-2023-0071.html), a
  timing sidechannel in exactly the private-key operation signing would use,
  still unpatched — while the party able to time our signature is the server we
  are authenticating to. pointycastle's RSA is not constant-time either, so this
  is not a claim that Dart is safer; it is that the swap buys nothing. A
  constant-time RSA means a BoringSSL-backed dependency and a C toolchain on
  five targets, which is a decision of its own. The seam leaves it open.
- **AES-GCM and ChaCha20-Poly1305.** The transport computes them inline rather
  than through the factories the seam sits behind. What is covered is what gets
  negotiated, since `SSHAlgorithms.cipher` proposes aes256-ctr first.
- **The file-transfer worker.** Its own isolate, so it would need its own
  `RustLib.init`, and it is not the isolate anyone is looking at.

Each is stated where the code is, so the next reader does not read them as
oversights.

## Also: private keys are parsed once, not once per connection

`genClient` ran `compute(loadIdentities, ...)` on every connection, so ten
servers sharing one key parsed it ten times in ten isolates — during launch,
where several connect at once. `identitiesOf` caches by the PEM rather than by
the id or path it came from, so replacing a key's contents parses the new one.
A key that failed to parse is cached too, or retrying it spawns an isolate per
connection to reach the same answer.

Sharing an `SSHKeyPair` is safe: every `sign` builds its own signer from the
key's fields and writes nothing back. Nor is a secret held that the process was
not already holding — `PrivateKeyUnlock` caches the decrypted PEM for the run,
and this is that PEM decoded.

## Testing

Signature verification is a different risk from a cipher: a wrong answer is a
forged host key accepted in silence, not a connection that drops. So the tests
are differential in both directions and adversarial.

- `crates/sbm_ffi`: 24 cases against NIST SP 800-38A, RFC 4231, RFC 7748,
  RFC 8032 and OpenBSD's bcrypt vectors, plus low-order point rejection and
  mpint width handling.
- `test/ssh_native_crypto_test.dart`: 48 cases. Each implementation must accept
  what the other signed, and both must refuse a flipped signature byte, a
  swapped `r` and `s`, a changed message and a changed key — across all three
  ECDSA curves with real keys. Ed25519 is compared byte for byte because it is
  deterministic; ECDSA is cross-verified because pointycastle's is randomised
  and the native one is RFC 6979. The record cipher is fed the way the transport
  feeds it — keyed once, then packet after packet, with the CTR counter and CBC
  chaining block carrying across calls.

`flutter analyze lib test integration_test` clean. `cargo test --workspace`: 28
suites pass. `flutter test`: 1822 pass, 1 skip, 1 fail —
`windows_install_ssh_e2e_test`, an opt-in e2e whose `setUpAll` cannot open
`~/.ssh/id_ed25519` with the passphrase in `.env`. It never installs the
backend, so nothing here reaches it; not reproduced against a clean tree.

## Not verified

Platform builds. Only `cargo build -p sbm_ffi` debug and release were run. The
new dependencies are pure Rust with no C, so cross-compilation should be
unaffected, but no five-platform build was done.

## Summary by CodeRabbit

- **New Features**
  - Added native SSH cryptography for modern ciphers, MACs, key exchange, signatures, and encrypted-key derivation.
  - Improved private-key handling with caching, concurrent parsing, clearer errors, and reliable cache invalidation.

- **Bug Fixes**
  - Invalid or unusable keys no longer prevent valid identities from being used.
  - Windows custom-command output now reads reliably up to the configured limit.

- **Chores**
  - Android release builds now include independent reproducibility verification before publication.

- **Documentation**
  - Updated English and Chinese guides covering SSH, Monitor agents, widgets, installation, architecture, building, and troubleshooting.


lollipopkit:
@coderabbitai review

PR #1395, opened 2026-09-04
feat(globe): see your servers on a globe, placed on this device

lollipopkit:
The server tab gains a globe. Each server is drawn where its address is, and **every coordinate is worked out on the device**: no address is sent to a geolocation service, and there is no per-lookup request that could disclose one.

Squashed from 31 commits on `feat/globe`, which is not pushed — it carried a 4 MB `assets/geo/country_v1.bin` that this branch adds and then deletes, and there was no reason to put it on GitHub permanently. The net diff is identical (verified: same tree).

## How a server is placed

The first of these that answers:

| Source | Needs the data? |
|---|---|
| A coordinate typed in the editor, under **More → Location** | no |
| A public address the machine reported for itself | yes |
| The city-level data, for the address the app connects at | yes |

The self-reported address comes from the ordinary status poll — `ip` is now a key in the shared command manifest, so SSH and a monitor agent answer it identically and neither needs a separate request. It exists for the machine behind NAT or reached over a VPN, where the address the app connects to and the address the world sees are not the same.

Anything unplaceable goes in a strip along the bottom that says **why**. "Unknown" over a tab of LAN servers is true and useless — that is the ordinary state of a homelab install — so a private address and an address no database covers are told apart.

## The data

**A 25 MB download, and nothing is bundled.** It is fetched once, after a dialog quoting this month's real sizes and DB-IP's attribution with a three-second confirm; every lookup after that reads a local file. The endpoint therefore learns nothing about which servers you have or when you look at them — a property of *having* the file rather than a promise about whoever serves it, and what makes this opt-in and off by default without a switch.

The 4 MB country database that used to ship with the app is gone. One point per country, and it cost every install megabytes to answer worse.

Distributed from `ipgeo.lollipopkit.com` with GitHub Releases as a fallback, built monthly by `lollipopkit/ipgeo-shards`. The format (`SBGX`) is pinned across both repositories by shared test vectors, because every way of getting it wrong is silent — a wrong stride does not fail, it answers.

## Also here

- **The crash report moved to Settings → Privacy.** It used to be a toast on the launch after a crash, raised once, and dismissing it lost the log for good. The log is never uploaded at any level; what `basic` and `full` send is the error. A crash from before the upload sink was installed — early startup, which nothing reported — is now replayed on the next launch, without double-filing the ones that were sent live.
- **Schema v20**: the per-host coordinate cache is dropped. A lookup is a handful of reads from an already-open file, so caching it cost more than the answer — and a cached entry never changed after a data update. What is still kept is the address a server reported for itself, which only that machine knows.
- The server's About card shows its public IP behind an eye toggle.
- `parse_ips` in `sbm_parser`, shared by both transports.
- Two new documentation pages per language, plus the privacy policy's account of what is downloaded and what stays on the device.

## Checks

- `flutter analyze lib test` clean; **2294 tests pass**. The one failure locally is `windows_install_ssh_e2e_test.dart`, which is environment-gated and needs `SBM_E2E_SSH_KEY_PASSPHRASE`.
- `packages/fl_lib` is at `5f45dc9`, already on its own `main`.
- Reviewed twice during development; 29 findings fixed, including a fd leak per lookup on a corrupt bundle, an IPv4-mapped address that could never be placed, and a download that emptied the globe for its own duration.

## Worth a careful look

- `m019_drop_geo_cache` deletes `kv` rows no *released* build ever wrote — the store was added and removed inside this branch. It stays because dev installs off this branch do have them, and `SchemaVersion.current` cannot go back to 19 without those installs refusing to open.
- The globe's card layout and leader lines have no pixel-level tests. The geometry is unit-tested; what a card looks like against the unplaced strip is not.

## Summary by CodeRabbit

* **New Features**
  * Added an interactive globe view for visualizing server locations, with rotation, zoom, selection, filtering, and responsive layouts.
  * Added automatic location resolution using manual coordinates, reported public addresses, and local city-level geographic data.
  * Added settings to enable the globe and download, update, or remove geographic data with consent and progress indicators.
  * Server details can display a concealed, revealable public IP address and support manually entered coordinates.
  * Added persistent crash-report viewing, copying, submission, and deletion.

* **Bug Fixes**
  * Improved cross-platform interface-address collection and preservation.
  * Installer uninstall prompts now safely handle piped installations.

* **Documentation**
  * Added globe, location, privacy, attribution, and geographic-data guidance in English and Chinese.


lollipopkit:
@coderabbitai review

Large change (13k lines) touching storage, a schema migration, and what leaves the device. Two internal review rounds already ran and their findings are fixed, so a third pair of eyes is mostly about what those missed.

Worth the most attention:
- `lib/core/service/geo_data.dart` and `geo_bundle.dart` — a format parsed from a downloaded file, where a wrong stride answers rather than failing.
- `lib/data/store/migrations/m019_drop_geo_cache.dart` — a `DELETE` on the `kv` table `setting` and `history` share.
- `lib/core/utils/private_address.dart` — the gate in front of every lookup; a range in the wrong class fails silently in both directions.

PR #1401, opened 2026-09-05
feat: benchmark a server with yabs

lollipopkit:
Runs [Yet Another Bench Script](https://github.com/masonr/yet-another-bench-script) from a page instead of a terminal — disk, network and CPU — and keeps the results structured rather than scrolled past.

## The script is shipped, not fetched

`curl -sL yabs.sh | bash` pins no version and verifies nothing, and `raw.githubusercontent.com` is unreachable from a good share of the hosts people most want to benchmark. `assets/yabs.sh` is vendored verbatim under WTFPL, pinned by commit and SHA-256, refreshed by `scripts/update-yabs.sh`, and held by a test.

## The run is detached; the app only watches it

A full run is 10–20 minutes — longer than a phone holds a connection or the OS leaves a backgrounded app alive. The launcher starts under `setsid` and the app polls a directory, so closing the page, locking the phone or losing the network costs nothing, and reopening picks the run back up from its row.

That also makes the transport irrelevant: every command is short, so this works over **SSH and over a monitor agent's `/exec`** with neither knowing about the other.

The launcher records its own `$$`, not the shell's `$!`. Under `setsid` that pid is the process group every child inherits, and killing the group is the only thing that actually stops a benchmark — fio, iperf3 and Geekbench are separate processes.

## Every phase is the user's choice

This supplies defaults and nothing more. Three disagree with yabs' own, each in the direction that spends less or discloses less:

| Option | Default | Why |
|---|---|---|
| Geekbench | **off** | downloads a proprietary binary and **publishes** CPU model, core count and memory to a public geekbench.com page |
| reduced iperf | **on** | seven locations, both directions, both address families is tens of GB of egress on a metered plan |
| IP lookup | **off** | plaintext HTTP to ip-api.com |

Estimated time, traffic and required free space sit above the Run button, because all three are invisible when the decision is made and expensive afterwards.

## Also here

- **`feat(monitor)`** — `/exec`'s 60s timeout and two 1 MiB caps become `[remote_access.exec]`. They were sized for listing processes and units, which is not a size anything taking minutes fits inside. Still the agent's decision; a request cannot raise them. Byte caps derive from RAM with the old flat value as the floor, so no machine gets less than it did.
- **`chore`** — `shaders/` moves under `assets/`. The `shaders:` pubspec section stays put: it is a compile step (impellerc → `IPLR`), not a copy, and Flutter refuses the combination outright.
- **`rm`** — the desktop-only refresh button in the server tab. Pull-to-refresh remains on phones; the poll runs on its own and each card refreshes its own server.

## Notes for review

- **Results parse leniently and the raw text is kept regardless.** yabs assembles its JSON by concatenating shell strings, so a field it could not collect arrives as an empty slot and a distro name containing a quote yields a document no parser accepts. Nulls mean "not measured", never zero.
- **`benchmark_run` is schema v21.** Not a sync root — a measurement is not an edit — and capped per server, except a `running` row is never pruned: it names a directory with a live process in it.
- Two values on these command lines are typed by the user, so `test/yabs_script_test.dart` runs the real fragments against `/bin/sh` and asserts an injected `; touch` does not execute, that `$HOME` expands, that cancelling takes the whole process group, and that a log cannot forge a state line.
- `test/benchmark_page_layout_test.dart` measures the rendered rows: three widget kinds share one column and each derives its left edge differently. Verified to fail on the two layouts it replaced.

## Status

`flutter analyze` clean · 2398 Dart tests pass (the one failure is a pre-existing opt-in SSH e2e needing `SBM_E2E_SSH_KEY_PASSPHRASE`) · monitor `cargo test` and `clippy` clean.

**Not yet run end to end against a real server** — the shell fragments are tested against a real `/bin/sh` with a stand-in for yabs, but a full 15-minute run on real hardware has not happened. Worth doing before this leaves draft.

## Summary by CodeRabbit

* **New Features**
  * Added server benchmarking for disk, network, CPU, and Geekbench tests.
  * Added live progress, cancellation, results, logs, comparisons, estimates, and history.
  * Added responsive benchmark layouts, server search and filtering, localized text, and server selection.
  * Added immersive globe viewing with action controls.
  * Added configurable command execution limits for the monitoring API.

* **Bug Fixes**
  * Improved benchmark polling, cleanup safety, result parsing, and history retention.
  * Improved globe labeling and layout around controls.
  * Removed custom iPerf server configuration.


lollipopkit:
@coderabbitai review

Points worth attention:

- `YabsScript` builds shell command lines that carry two user-typed values (working directory, custom iperf server list). `quote`/`quotePath` are the only thing between those and a shell on someone else's machine.
- `cleanupCommand` issues `rm -rf` on a path derived from user input; it asserts the path shape first.
- The detached-run lifecycle: start → poll → finish/cancel/cleanup in `lib/data/provider/benchmark.dart`, especially `YabsPollState.answered` (an unrecognised answer must not be read as "the run is gone").
- Schema v21: the three edits (table, `SchemaVersion.current`, `kSchemaMigrations`) and whether m020's DDL matches what Drift creates.
- `monitor/src/core/remote_access.rs` — the derived exec bounds and whether the floor really prevents a regression on small machines.

lollipopkit:
Worked through the review. Four fixed, two skipped with reasons.

## Fixed

**`benchmark.dart:130` — a run failed before its launcher started.** Real, and it would have hit nearly every run. `startEntry` creates the directory and returns as soon as it has backgrounded the launcher; the launcher writes its pid only once scheduled. With `_schedule(immediate: true)` the first poll fires at `Duration.zero` and saw *directory + no process + no exit code* — identical to an OOM kill. The poll now reports whether the pid file exists, and only a run that got that far can be called dead; a launcher that never appears is bounded by a 30s grace period. `31146470`

**`benchmark.dart:260` — cleanup ownership.** Took the marker suggestion: `startEntry` stamps the run id into `owner` before detaching, and `cleanupCommand` requires it to match. Kept the shape check alongside it — the shape rules out a path this code could not have produced, the marker rules out a directory of that shape belonging to something else. One marker serves both findings. `31146470`

**`m020:30` — partial migration.** Real. The version is recorded only after `apply()` returns, so a process that stopped between the table and its index was *guaranteed* to take the early return on the next launch and leave the index missing forever. Each statement guards itself now. `f507915d`

**`update-yabs.sh:28` — clobbered the asset on failure.** Real, and the version check that would have caught it ran after the damage. Downloads beside the destination, moves into place only once the file identifies itself. `f507915d`

## Skipped

**`yabs_script_test.dart:229` — "extra closing `fi`".** Not present. `cancelCommand` has one `if` and one `fi`. `test/yabs_script_test.dart`'s cancel test runs that exact script through a real `/bin/sh`, kills a launcher plus a child it spawned, and asserts exit 143 — a stray `fi` is a syntax error and could not pass.

**`globe.frag:82` — premultiplied alpha.** Not reachable, and this PR does not touch the file (`{shaders => assets/shaders}/globe.frag | 0`, a pure rename). Both colours feeding `surface` are `Color.alphaBlend(fg, <opaque theme colour>)`, so their alpha is 1.0 and premultiplied equals straight. The one translucent uniform is already premultiplied correctly: `vec4(uGlow.rgb, 1.0) * uGlow.a * halo`.

## Also in this push

`6690aa46` fixes an alignment bug the screenshots caught: the two text-field rows used their fields' `InputDecoration` labels as row names, and `InputDecorator` paints a resting label through a transform of its own — so they sat ~6 logical px right of every other title. `getTopLeft` reported all of them at the same 74.0, which is why the layout test passed; measured on the rendered image the switch titles are at 113-115px and the two labels at 122 and 125. The name is an ordinary `Text` in the shared column now, so alignment is structural.

## Still open

Not yet run end to end against a real server. And `test/yabs_script_test.dart`'s `runToCompletion` has timed out twice under a parallel suite run and did not reproduce in three subsequent full runs — root cause not established, so its timeout now dumps the run directory, the generated `run.sh` and the last poll output.

PR #1403, opened 2026-09-05
fix: drop snap images, swap areas and container layers from the disk list

lollipopkit:
Reported with a screenshot: a server's **Devices** list was eight `/dev/loopN (/snap/...)` rows at 100% before the machine's actual disks appeared.

## What was wrong

A snap-heavy Ubuntu mounts one squashfs per installed revision. Each is full by construction, so besides filling the list, every one added its whole size to the disk total — the reported machine was off by more than a gigabyte of phantom "used".

Two more classes turned up while checking what else survives the filter. I fed constructed `df` and `lsblk` output through the parser rather than guessing:

| Class | Behaviour | Path |
|---|---|---|
| Read-only images | snap squashfs, `snapfuse`, mounted ISO — 100% full, counted in the total | `lsblk` and `df` |
| Swap areas | `lsblk` lists them beside filesystems with an empty `FSSIZE`, rendering as `0 B / 0 B` | `lsblk` |
| `fuse-overlayfs` | rootless podman/docker publishes a row per container carrying the host filesystem's own numbers | `df` fallback |

## The rules

`disk_should_calc` (`crates/sbm_parser/src/types.rs`) excludes all three before deciding, alongside the existing kernel-mount check.

Read-only images are matched **on the image, never on `/dev/loop`**. A loop device carrying a writable filesystem is storage someone mounted on purpose, and the `df` fallback hosts — busybox, OpenWrt, no `lsblk` — are exactly where that is most likely. `/dev/loop7 → /mnt/image` with ext4 stays visible, and there is a test for it.

Swap already has its own reading from `/proc/meminfo`, so the block-device row was duplicate information with a broken number.

Container layers are matched by exact filesystem name — `overlay`, `overlayfs`, `fuse-overlayfs`. The first two are docker's own layers in their current and pre-4.0 spellings; the third is what a rootless podman or docker mounts.

These were prefix matches until review (`overlay` covering `overlayfs` for free). A `df` source is user-chosen text, so a prefix reached too far: a ZFS pool named `tmpfspool`, or an export from an NFS host named `shm-nas`, each begin with one of these names and each is storage. See the second commit.

## Dart side

The parser drops these while reading `lsblk`/`df`, but a `Disk` built from a monitor agent's `disk_details` never went through it — an agent older than a given rule still sends what that rule removes. So the rules are mirrored in `lib/data/model/server/disk.dart` and applied in `monitor_metrics_mapper.dart`.

The call sites now ask one thing, `Disk.isStorage`, which runs `_shouldCalc` against both the source and the filesystem type. The parser is handed one or the other and branches differently for each; a parsed `Disk` carries both.

## Known limit

`df -k` reports no filesystem type, so a mounted ISO reads as an ordinary `/dev/sr0` there and is not dropped. It *is* dropped under `lsblk`, which is every host that has an optical drive; `df` is the fallback for busybox and OpenWrt. Recognising `/dev/sr` by name would trade a rule about images for a list of device names, so `disk_parse_df_drops_read_only_images` records the behaviour with a positive assertion instead — if someone changes it, that test says so.

## Checked and deliberately not changed

Verified by running the parser, so these need no rule:

- `ramfs` (systemd credential mounts), `nsfs`, `gvfsd-fuse` — `df` reports `-` for usage, which fails to parse and drops the row already. Confirmed both spellings of systemd's credential mount on an orbstack NixOS machine (tmpfs there) and against the ramfs form.
- docker bind volumes, btrfs subvolumes — same source and same numbers, already collapsed by `parse_df`'s dedup.
- cgroup2, efivarfs, binfmt_misc, hugetlbfs, mqueue, pstore, bpf — all mounted under `/sys`, `/proc` or `/dev`.

Left visible on purpose: `/boot` and `/boot/efi`, LVM and `/dev/mapper/*`, NFS/CIFS/SSHFS, everything under `/mnt`.

Not done, as agreed: ZFS snapshot auto-mounts (`/.zfs/snapshot/`) need `snapdir=visible` to appear at all, and flatpak's `portal` is desktop-only.

## Tests

`crates/sbm_parser/tests/fixtures/lsblk.json` came off a real machine and carries that machine's swap partition — `disk_parse_lsblk_json` had been asserting 6 entries with swap as one of them. It is 5 now, which is the clearest evidence this was not a constructed problem.

Four new parser tests over two new fixtures (`lsblk_nonstorage.json`, `df_nonstorage.txt`), plus three `DiskUsage` cases and one mapper case on the Dart side.

- `cargo test -p sbm_parser` — 88 + 45 pass
- `cargo check --workspace --all-targets` — clean
- `flutter test test/disk_test.dart test/monitor_metrics_test.dart test/server_status_update_req_test.dart` — pass
- `flutter analyze` on the touched files — clean

## Note for reviewers

The filter is in Rust, so a hot reload will not pick it up — this needs an app rebuild to see.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved disk usage reporting by excluding non-storage entries such as read-only images, swap areas, snap mounts, container overlays, kernel mounts, ramfs, and other virtual filesystems.
  * Preserved writable loop-backed filesystems and storage sources whose names resemble virtual filesystem prefixes.
  * Ensured disk details and aggregate usage remain accurate when legacy payloads include image-based mounts.

* **Tests**
  * Added coverage for non-storage filtering and accurate disk usage totals across supported filesystem types.

### DIFF
--- a/crates/sbm_parser/src/types.rs
+++ b/crates/sbm_parser/src/types.rs
@@ -151,10 +151,15 @@ pub fn disk_should_calc(fs: &str, mount: &str) -> bool {
 /// `overlay` and `overlayfs` are the current and pre-4.0 spellings of docker's
 /// own layers. `fuse-overlayfs` is what a rootless podman or docker mounts,
 /// one row per container carrying the host filesystem's own numbers.
+///
+/// `ramfs` is systemd's credential mounts, one per unit that takes a
+/// credential. Most `df` builds report `-` for its usage, which fails to parse
+/// and drops the row before any of this; naming it here does not depend on
+/// that, since a build reporting `0%` instead leaves a row of 0 B.
 fn is_virtual_fs(fs: &str) -> bool {
     matches!(
         fs,
-        "shm" | "tmpfs" | "devtmpfs" | "overlay" | "overlayfs" | "fuse-overlayfs"
+        "shm" | "tmpfs" | "ramfs" | "devtmpfs" | "overlay" | "overlayfs" | "fuse-overlayfs"
     )
 }
 
--- a/crates/sbm_parser/src/types.rs
+++ b/crates/sbm_parser/src/types.rs
@@ -121,23 +121,55 @@ pub struct Disk {
 }
 
 /// Whether a `df` row describes storage worth reporting (Dart `_shouldCalc`):
-/// exclude kernel mounts; include /dev-prefixed sources, network mounts (//),
-/// /mnt mount points; exclude shm/overlay/tmpfs/devtmpfs sources; include the rest
+/// exclude kernel mounts, read-only images and swap areas; include /dev-prefixed
+/// sources, network mounts (//), /mnt mount points; exclude
+/// shm/overlay/fuse-overlayfs/tmpfs/devtmpfs sources; include the rest
 pub fn disk_should_calc(fs: &str, mount: &str) -> bool {
     // Checked before the inclusions below, so that a source spelled like a
     // block device cannot bring these back: a host that exposes many device
     // nodes publishes one devtmpfs row per node, two dozen lines all claiming
     // 0 B used of the same size.
-    if is_kernel_mount(mount) {
+    if is_kernel_mount(mount) || is_read_only_image(fs, mount) || is_swap_area(fs, mount) {
         return false;
     }
     if fs.starts_with("/dev") || fs.starts_with("//") || mount.starts_with("/mnt") {
         return true;
     }
+    // `overlay` covers the old `overlayfs` spelling too, but not the
+    // `fuse-overlayfs` a rootless podman or docker uses: that one publishes a
+    // row per container carrying the host filesystem's own numbers.
     !(fs.starts_with("shm")
         || fs.starts_with("overlay")
+        || fs == "fuse-overlayfs"
         || fs.starts_with("tmpfs")
         || fs.starts_with("devtmpfs"))
+}
+
+/// A read-only image mounted as a filesystem — a snap's squashfs, the same
+/// image handed to a container through `snapfuse`, a mounted ISO — rather than
+/// storage anyone manages. Each is full by construction, so it reports 100%
+/// and contributes its whole size to the total; a snap-heavy Ubuntu publishes
+/// twenty of them, which is the entire device list.
+///
+/// Matched on the image, never on `/dev/loop`: a loop device carrying a
+/// writable filesystem is storage someone mounted on purpose, and it is the
+/// `df` fallback hosts (busybox, no lsblk) where that is most likely.
+/// `fs` is a filesystem type under `lsblk` and a source under `df`, so the
+/// two sources are recognised by different halves of this.
+fn is_read_only_image(fs: &str, mount: &str) -> bool {
+    matches!(
+        fs,
+        "squashfs" | "erofs" | "iso9660" | "snapfuse" | "fuse.snapfuse"
+    ) || mount.starts_with("/snap/")
+        || mount.starts_with("/var/lib/snapd/snap/")
+}
+
+/// A swap area, which `lsblk` lists beside filesystems. It has no mount point
+/// and no `FSSIZE`, so the row reads `0 B / 0 B`, and swap is reported on its
+/// own from `/proc/meminfo` anyway. `df` lists only mounted filesystems and
+/// never produces one of these.
+fn is_swap_area(fs: &str, mount: &str) -> bool {
+    fs == "swap" || mount == "[SWAP]"
 }
 
 /// `/dev`, `/proc`, `/sys`, and anything mounted inside them.
--- a/crates/sbm_parser/src/types.rs
+++ b/crates/sbm_parser/src/types.rs
@@ -122,8 +122,11 @@ pub struct Disk {
 
 /// Whether a `df` row describes storage worth reporting (Dart `_shouldCalc`):
 /// exclude kernel mounts, read-only images and swap areas; include /dev-prefixed
-/// sources, network mounts (//), /mnt mount points; exclude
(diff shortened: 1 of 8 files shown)


lollipopkit:
@coderabbit review

Points worth the most attention:

1. **False positives.** The three new rules exclude rows from the disk list and from the total. Is there a real filesystem that matches `squashfs`/`erofs`/`iso9660`/`snapfuse`/`fuse.snapfuse`, `swap`/`[SWAP]`, or `fuse-overlayfs` and *is* storage a user manages? A wrong exclusion here is silent — the disk simply stops existing in the UI.

2. **The overloaded first argument.** `disk_should_calc(fs, mount)` receives a filesystem *type* from the `lsblk` branch and a *source* from the `df` branch. Both new predicates are written against that. Check the `df` branch cannot pass a source that collides with one of the type strings, and vice versa.

3. **Rust/Dart drift.** `is_read_only_image` / `is_swap_area` in `crates/sbm_parser/src/types.rs` are mirrored by `_isReadOnlyImage` / `_isSwapArea` in `lib/data/model/server/disk.dart`. They must stay identical. Is there anything in the Dart mirror that diverges?

4. **`Disk.isStorage` calls `_shouldCalc` twice**, once with `path` and once with `fsTyp ?? path`. Confirm the conjunction cannot reject a legitimate disk that the single-argument parser would keep — particularly a row where `fsTyp` is null and `path` is not a `/dev` device (NFS, CIFS).

lollipopkit:
Both findings were valid. Fixed in 2b9aca81.

**Prefix matching on virtual filesystems.** Confirmed by running the parser: a `df` source is user-chosen text, so `startsWith` reached past the filesystem names. `tmpfspool/data` (a legal ZFS pool name), `shm-nas:/vol1`, `overlay-01:/export` and `devtmpfs-backup:/snapshots` were each dropped from the device list *and* from the disk total. Now matched exactly, via `is_virtual_fs` / `_isVirtualFs`, with `overlayfs` and `fuse-overlayfs` spelled out where the `overlay` prefix used to cover the first and a separate equality check the second. Applied to both the Rust rule and the Dart mirror — the finding named only the Dart side, but they have to stay identical.

**`disk_usage` asking with the path alone.** This one was worse than described, and it was mine: the previous commit changed the Dart `DiskUsage.parse` to ask with both identifiers and left the Rust `disk_usage` asking with one. `disk_should_calc` takes *one* of a row's two identifiers and branches differently for each, so a row identified only by its filesystem type was missed — a squashfs mounted outside `/snap`, or a swap area whose mount `lsblk` left empty rather than `[SWAP]`. It now goes through a new `Disk::is_storage`, mirroring `Disk.isStorage`.

Worth stating plainly: **no production code calls `disk_usage`** — the app computes this in Dart, and I checked `crates/` and `monitor/src/` for callers. The reason to fix it anyway is that `dart_compat.rs` is the file that asserts the two implementations agree, and left diverging it would have gone on asserting a parity that no longer held.

**On the requested regression tests.** Added `disk_parse_df_keeps_storage_named_like_a_virtual_fs` (the four sources above), `disk_usage_drops_rows_identified_only_by_filesystem_type`, and a Dart `DiskUsage keeps storage named like a virtual filesystem`. I then reverted both rules in place and re-ran: both new Rust tests fail against the old code, and pass against the new. A test that would have passed either way is not a regression test.

`cargo test -p sbm_parser` 90 + 45 pass, `cargo check --workspace --all-targets` clean, `flutter test` on the three touched suites passes, `flutter analyze` clean.

Not changed: the exact-match set is the six kernel-backed names, not a broader list. `ramfs` and `nsfs` are still handled by `df` reporting `-` for usage, which fails to parse and drops the row — verified, and recorded in `disk_parse_df_drops_container_layers`.

lollipopkit:
One of the two applied. Fixed in 9d1af534.

## `ramfs` — valid, done

Verified rather than assumed: I fed the parser a `df` whose ramfs row reports a parseable `0%` instead of the `-` most builds print, and the row survives as 0 B. The existing protection was the usage column failing to parse, which is an accident of formatting, not a rule. `ramfs` is now named in `is_virtual_fs` / `_isVirtualFs`, exact match — an NFS export from a host named `ramfs-nas` is unaffected.

Covered by `disk_parse_df_drops_ramfs_reporting_a_usage`. The comment on `disk_parse_df_drops_container_layers` said ramfs "needs no rule"; that was true for the `-` shape only, and now points at the new test.

## `is_virtual_fs` ahead of the `/mnt` rule in `is_storage` — skipped

The behaviour described is real: `disk_should_calc` returns early on `/dev`, `//` and `/mnt` before reaching the virtual-filesystem check, so a tmpfs mounted at `/mnt/ramdisk` is treated as storage. I reproduced it — it comes back as a 1 GiB disk.

Two reasons not to take the change as proposed:

1. **It would create the inconsistency it is meant to remove.** The suggestion is to reorder inside `Disk::is_storage` only. Parsing uses `disk_should_calc` directly (`linux.rs` lines 106, 146, 161, 228), so the row would stay in the device list and disappear from the total. A 1 GiB disk shown in the list and not counted is worse than the current state, where list and total agree.

2. **It is not this class of row, and not this PR.** The `/mnt` precedence predates the PR and is untouched by it. What this PR hides is rows that are full by construction (read-only images) or report 0 B (swap, ramfs). A tmpfs ramdisk has a real capacity and a real usage that moves — someone who ran `mount -t tmpfs tmpfs /mnt/ramdisk` plausibly wants to watch it. Whether `/mnt` should keep outranking the virtual-filesystem check is a genuine question, but it is a behaviour change with a user-visible loss, and it belongs in its own change rather than riding along here.

Happy to open an issue for it if you would like it tracked.

`cargo test -p sbm_parser` 91 + 45 pass, `cargo check --workspace --all-targets` clean, `flutter test` and `flutter analyze` clean.

PR #1408, opened 2026-09-06
fix(ios): ship yabs as base64, not a bundled shell script

lollipopkit:
Three things, all found while working out why the v1574 iOS upload was rejected.

## The upload failure

App Store validation walks everything inside `Runner.app` and treats a file it reads as executable code — a script suffix, or a leading `#!` — as a nested code object that must carry its own signature. Nothing under `flutter_assets` does. So `assets/yabs.sh` cost nothing at build time and failed the *upload*:

> Invalid Signature. Code object is not signed at all. The file at path "Runner.app/Frameworks/App.framework/flutter_assets/assets/yabs.sh" is not properly signed.

— an error that names the certificates and not the file.

The asset is base64 now (`assets/yabs.b64`). Both things the scanner looks at are gone: the `.sh` suffix and the leading `#!/bin/bash`. That is why this is an encoding and not a rename — a `.txt` still opens with a shebang.

- `YabsScript.decodeAsset` is the single decoder, shared with the test.
- `scripts/update-yabs.sh` writes the asset with `openssl base64` (`base64`'s wrapping flag is `-w` on GNU and `-b` on BSD), and takes the digest over the script it downloaded, never over the encoding.
- `sha256Hex` is unchanged at `c42397c6…` and still means what it did: the exact program a server is sent.

The regression test walks every entry of pubspec's `assets:` with directories expanded, checking the suffix and the first two bytes, since the next asset to do this will not be this one. Verified it fails when a `.sh` is dropped into `assets/geo/`.

## monitor v0.1.2

Its version is decoupled from the app's and read from the crate by `monitor-release.yml`, so a release needs the bump in the tree first. Since `monitor-v0.1.1` (2026-08-27):

| Change | Effect |
|---|---|
| `ips` in `SystemMetrics` (#1395) | The app reads it from `/api/v1/metrics` to place a server on the globe. A 0.1.1 agent does not send it. |
| `[remote_access.exec]` (#1401) | The 60s timeout and the two 1 MiB caps were constants; a benchmark or a filesystem scan had no way to ask for more. |
| `sbm_parser` disk fixes (#1403, #1404) | The agent samples through `sbm_native`, so its disk list still carries snap images, swap and container layers. |
| russh 0.63, `ratelimit.rs` (#1369) | Dependency and implementation updates. |

No new routes, so an older app keeps working against a newer agent.

## File name

`scripts/release/prune-android-dev-plugins.dart` was the project's only `file_names` violation, and the one other Dart file beside it is already snake_case. Both F-Droid scripts name it by path and are what the reproducible-build workflow runs, so the rename and their call sites move together.

## Verification

- `flutter test test/yabs_script_test.dart` — 27 pass
- `flutter test test/prune_android_dev_plugins_test.dart` — passes
- `flutter analyze` over the whole project — no issues
- `flutter build bundle` — `build/flutter_assets` contains no file with a script suffix

Not verified: an actual App Store upload. That needs a build number past 1574, which is a separate bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved benchmark asset handling to support reliable loading and validation while preserving benchmark functionality.
  * Added safeguards to prevent bundled assets from being treated as executable scripts during app validation.

* **Chores**
  * Updated Android release tooling and plugin cleanup processes.
  * Updated the monitor package version to 0.1.2.

* **Tests**
  * Expanded coverage for benchmark asset decoding, integrity verification, and Android plugin cleanup.


lollipopkit:
@coderabbit review

PR #1409, opened 2026-09-06
opt.: cut the monitor binary by a third

lollipopkit:
22.0 MB → 15.2 MB, measured on macOS arm64 with the workspace's existing `strip`. For reference, the `linux_amd64` archive published at v0.1.2 is 10.65 MB.

| Build | Binary |
|---|---|
| v0.1.2, as shipped | 21,983,232 B |
| without aws-lc | 20,099,360 B |
| `+ release-monitor` profile | 15,222,016 B |
| *(`opt-level = "s"`, not taken)* | *11,159,072 B* |

## Two crypto implementations were linked in

`cargo tree -i aws-lc-sys` found aws-lc-rs beside the `ring` that rustls, jsonwebtoken and russh are each configured with by hand. reqwest brought it: its `rustls` feature is defined as `__rustls-aws-lc-rs`, so the provider arrives whatever the direct rustls dependency asks for.

`monitor/Cargo.toml` has described this since it was written, including the fix — `rustls-no-provider` plus a process-level provider installed before the client is built. That is done here, in `push::http_client`, where the client is created.

Without the install, `Client::builder().build()` panics inside reqwest rather than failing as a push error. Verified by removing the line: the four webhook tests in `monitoring::push` panic at `reqwest-0.13.4/src/async_impl/client.rs:2484`. Those tests are what will catch a future TLS dependency bringing its own provider back.

−1.8 MB, and cmake leaves the musl build — the stated reason those features were chosen in the first place.

## `release-monitor`

`release` plus `lto = "fat"` and `codegen-units = 1`. Both are profile-wide — `[profile.release.package.*]` accepts neither — so asking for them on the shared profile would put a fat-LTO link in front of every app build, for a binary the app does not ship. `monitor-release.yml` is the only consumer.

−4.9 MB, at about +1 minute per target.

`opt-level = "s"` reaches 11.2 MB and is deliberately not taken: this process samples the machine and terminates every TLS connection to it. The number is recorded beside the profile so it stays a decision rather than an oversight.

## `monitor/Dockerfile` was shipping an unstripped binary

It rewrites the `sbm_parser` path, which makes `monitor/Cargo.toml` the root — so the build is outside the workspace and outside every `[profile]` in it, `strip` included. It now passes the same three settings as `CARGO_PROFILE_RELEASE_*`. A profile section in that manifest is not an option: cargo ignores a non-root package's profiles with a warning, and the workspace also reads that file.

`nix/package.nix` builds inside the workspace, so it keeps `strip` and does not get the new profile. Moving it needs `cargoBuildType`, whose handling differs between nixpkgs releases and is not verified here — noted in `monitor/CLAUDE.md` rather than guessed at.

## Verification

- `cargo test -p server_box_monitor` — 160 unit tests plus every integration test, all passing
- `cargo build -p server_box_monitor --profile release-monitor` — lands in `target/release-monitor/`, which is what the workflow's two path references now point at
- `cargo tree -i aws-lc-sys` — no match
- Not verified: an actual `monitor/Dockerfile` image build, and the musl numbers (the measurements above are macOS arm64)

The next `monitor-release.yml` run needs a version bump first; 0.1.2 is published and the workflow refuses an existing tag.

## Summary by CodeRabbit

- **Improvements**
  - Monitor release builds now use dedicated optimization settings to produce smaller, more efficient artifacts.
  - Monitor packaging and container builds consistently use the optimized monitor build output.
  - Monitoring HTTPS requests now use the Ring cryptography provider for consistent TLS behavior.
  - GitHub Release notes focus on essential Docker, checksum, and platform information.
  - Test cleanup is more reliable when background benchmark processes temporarily hold files.

- **Documentation**
  - Added guidance on monitor build profiles, optimization settings, container behavior, TLS requirements, and expected binary sizes.

---

## Also here: the release notes were wrapped at 72 columns

GitHub renders release notes with hard line breaks on, so the paragraphs wrapped in the workflow's `body:` came out wrapped on the page too — [v0.1.2](https://github.com/lollipopkit/flutter_server_box/releases/tag/monitor-v0.1.2) published as a column of short lines down the left of a full-width screen.

One line per paragraph now, however long. The fenced blocks keep their own breaks, which are the commands'. v0.1.2's notes have been rewritten in place, so the fix is only for the next release.

### DIFF
--- a/.github/workflows/monitor-release.yml
+++ b/.github/workflows/monitor-release.yml
@@ -141,9 +141,13 @@ jobs:
         with:
           key: ${{ matrix.target }}
 
+      # `release-monitor`, not `release`: fat LTO and one codegen unit, which
+      # are profile-wide settings and so cannot be asked for on the shared
+      # profile without slowing every app build. Defined in the workspace
+      # Cargo.toml, which says what it costs and what it buys.
       - name: Build
         shell: bash
-        run: cargo build -p server_box_monitor --release --target ${{ matrix.target }}
+        run: cargo build -p server_box_monitor --profile release-monitor --target ${{ matrix.target }}
 
       - name: Download frontend dist
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
@@ -162,7 +166,7 @@ jobs:
 
           staging="${BIN_NAME}"
           mkdir -p "$staging"
-          cp "target/${{ matrix.target }}/release/${bin}${ext}" "$staging/"
+          cp "target/${{ matrix.target }}/release-monitor/${bin}${ext}" "$staging/"
           mkdir -p "$staging/frontend"
           cp -r monitor/frontend/dist "$staging/frontend/dist"
           cp -r monitor/migrations "$staging/migrations"
@@ -207,7 +211,7 @@ jobs:
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
           name: monitor-bin-linux-${{ matrix.arch }}
-          path: target/${{ matrix.target }}/release/server_box_monitor
+          path: target/${{ matrix.target }}/release-monitor/server_box_monitor
           if-no-files-found: error
           retention-days: 3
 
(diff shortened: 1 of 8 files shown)


lollipopkit:
@coderabbit review

PR #1441, opened 2026-09-07
chore(deps): point fl_lib at its merged commit

lollipopkit:
lollipopkit/fl_lib#51 was squashed onto that repository's main as `ee1c720`. The gitlink still named `0024513` — the PR branch's head, which after the branch is deleted is reachable from nothing on main. Every checkout in this repository uses `submodules: recursive`, so that is the whole CI, not one job.

The two commits have an identical tree (`git diff ee1c7200 0024513` is empty), so this changes no code; it only re-anchors the pointer onto a commit main can reach.

Merging this is what makes it safe to delete `feat/update-machine-arch` in fl_lib.

## Summary by CodeRabbit

- **Chores**
  - Updated the underlying shared library version.


lollipopkit:
Folding into #1440 instead — same file-level unrelatedness, one fewer PR to shepherd.

PR #1444, opened 2026-09-08
fix(monitor): bound the agent's SQLite I/O and the history window

lollipopkit:
Three defects found while tracing an OrbStack Helper memory question. They turned out not to be the cause of that — the helper's growth is reclaimable page cache of the VM disk image and it falls back — but each is real on its own, and together they are the agent's continuous disk I/O.

Measured on an agent running for a month, with no client connected:

| | before |
|---|---|
| disk read caused by the agent | 431 KB/s, 0 with the agent stopped |
| read to append one ~150-byte row | 4,280,320 bytes in one cycle, cold cache |
| write per 7-second cycle | 53–74 KB |
| database file | 136 MB, 20329 of 33331 pages on the freelist |
| `/metrics/history?minutes=60` | 4255 rows scanned, 496 in the window |

## Connection settings (`db/database.rs`)

They move onto `SqliteConnectOptions` so the pool applies them to every connection. `synchronous` and `wal_autocheckpoint` are per-connection settings, and a pool opens connections lazily and replaces them over their lifetime, so a `PRAGMA` executed once against the pool configures whichever single connection answered it.

- `wal_autocheckpoint` 1000 → 128 pages. The agent commits a small row every few seconds forever, so the WAL sat near its 4 MB ceiling; rebuilding the WAL index after a page-cache eviction reads every frame, which is the 4 MB above. Checkpointing more often moves the same bytes into the database file in smaller pieces.
- `synchronous` stays at FULL, now named explicitly so that lowering it is a decision rather than an omission. It was NORMAL in the first push; the saving was never measured on its own, and this database also holds the panel credential, the watch tokens a revocation deletes from, and two audit logs that exist to survive the kind of event an unsynced commit is lost to.
- `journal_mode` and `auto_vacuum` named explicitly. sqlx 0.9 no longer sets a journal mode (`// Don't set journal_mode unless the user requested it`), and it is a persistent property of the file — existing databases stay in WAL, a newly created one would have got a rollback journal.
- `Sqlite::create_database` is no longer called. It opens a connection with default settings, and the file it leaves behind is past the point where `auto_vacuum` can still be chosen; `create_if_missing` makes the first connection to a new file one that carries every setting. Found by the test, which reported `left: 0, right: 2`.

## Space reclamation (`db/cleanup.rs`)

VACUUM ran on a timer fixed at seven cleanup intervals and starting a full period out, so a default install rewrote the whole file every seven days whether or not there was anything to reclaim — and an agent restarted more often than that reclaimed nothing, ever. `reclaim_free_pages` asks the freelist at the end of each cleanup pass, and acts only above 25% of the file and 1024 pages: `incremental_vacuum` when the database was created for it, one full VACUUM otherwise, which also converts it, since VACUUM adopts the running connection's `auto_vacuum`.

Existing agents therefore reclaim on their first cleanup pass after upgrading and use the incremental path from then on.

## History window (`api/server.rs`)

The bound was `datetime('now', '-N minutes')`, output `2026-09-08 07:28:22`, compared as text against a column holding `2026-09-08T08:28:18.493098803+00:00`. `'T'` (0x54) sorts above `' '` (0x20), so the comparison only ever distinguished the date and every row from the boundary's day passed it. The factor is "seconds into the UTC day / 3600", reaching 24x before UTC midnight.

The response was unaffected: the handler keeps the newest `max_points` buckets, so the extra rows were work and nothing else. Binding a `DateTime<Utc>` gives the query the encoding `store_metrics` writes the column with.

## Tests

`tests/db_pragmas.rs`, new: every pooled connection carries the settings (four connections held open at once, since acquiring them in sequence returns the same one), a new database is laid out for incremental reclaim, reclaim gives back the pages a deletion freed, a small freelist is left alone.

`tests/metrics_history_points.rs`: the fixture wrote timestamps with `datetime('now', ...)`, a shape nothing in the agent produces — it agreed with the query and with nothing that ships, which is how the bound stayed wrong with every test in the file passing. It now binds a `DateTime<Utc>` the way `store_metrics` does. `a_row_from_earlier_today_is_outside_a_five_minute_window` is the regression test, checked against the old predicate to confirm it fails there. It skips, with a printed reason, during the first five minutes of a UTC day, when "earlier today" is inside the window and the defect has nothing to show.

`cargo test` for the monitor: 283 passing, 1 pre-existing ignored. clippy clean.

## `access_log` timestamp (`migrations/009`, `api/ws/audit.rs`)

Same class of mismatch, found while checking whether the history bound was the only one. `access_log.timestamp` was `DATETIME DEFAULT CURRENT_TIMESTAMP` — `2026-09-08 08:28:18` — while `cleanup_policy_tables` compares it against a bound `DateTime<Utc>`:

```
sqlite> SELECT '2026-06-10 16:45:23' < '2026-06-10T16:45:22.123+00:00';
1
sqlite> SELECT '2026-06-10T16:45:23+00:00' < '2026-06-10T16:45:22.123+00:00';
0
```

A row one second *inside* the retention window read as older than the cutoff, so up to a day of audit log past the cutoff was collected on each pass.

Migration 009 rebuilds the table with `timestamp TEXT NOT NULL` and converts the existing rows; `Event::record` now supplies the value. The default is dropped rather than corrected — `NOT NULL` with nothing to fall back on makes a writer that forgets the column fail at the insert, where a default producing the right text would be a second encoder to keep in step with chrono's. Two test fixtures were relying on that default and failed exactly that way.

No foreign key points at this table or out of it, so the rebuild needs no `foreign_keys` or `legacy_alter_table` handling. Nothing reads the column but retention, so the stored shape reaches no client. `the_audit_log_cutoff_is_a_moment_not_a_date` asserts one second either side of the boundary; migration 009's checksum is pinned in `shipped_migrations_keep_their_checksums`. `migration_009_converts_the_timestamps_already_in_access_log` runs the migration against a database that genuinely predates it — `migrator_through(8)` builds a `Migrator` from a prefix of the same `Migration` values, and the row is written through the pre-009 default rather than typed by the fixture. It asserts the converted text against the text that was there, plus the other columns, the index and AUTOINCREMENT across the rebuild, and was confirmed to fail with the conversion changed back to the old shape.

`users.last_login` and `retention_policies.last_cleanup` also use `CURRENT_TIMESTAMP` and are left alone: both are written and never compared, so the shape means nothing there.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed metrics history filtering so results are limited to the requested time window.
  - Corrected audit-log retention to use precise timestamps, preventing premature deletion of recent entries.
  - Standardized timestamp storage for consistent sorting and retention behavior.

- **Performance & Reliability**
  - Database cleanup now automatically reclaims unused storage.
  - Improved database initialization and connection reliability.
  - Reduced unnecessary database scanning during metrics queries.

### DIFF
--- a/monitor/migrations/009_access_log_timestamp.sql
+++ b/monitor/migrations/009_access_log_timestamp.sql
@@ -1,0 +1,56 @@
+-- `access_log.timestamp` in the one text shape the rest of this database uses.
+--
+-- It was `DATETIME DEFAULT CURRENT_TIMESTAMP`, which writes
+-- `2026-09-08 08:28:18`, while every timestamp the agent binds itself is a
+-- `DateTime<Utc>` that sqlx encodes as RFC 3339
+-- (`2026-09-08T08:28:18.493098803+00:00`). Retention compares this column
+-- against a bound cutoff, SQLite compares the two as text, and `'T'` (0x54)
+-- sorts above `' '` (0x20) — so the comparison only ever distinguished the
+-- date, and every row from the cutoff's own day read as older than the cutoff
+-- whatever its time. Up to a day of audit log past the cutoff was deleted on
+-- each cleanup pass.
+--
+-- The default goes with it rather than being corrected. `NOT NULL` with
+-- nothing to fall back on means a writer that forgets the column fails at the
+-- insert, which is the failure this is trying not to have again; a default
+-- that happens to produce the right text is a second encoder to keep in step
+-- with chrono's.
+--
+-- No foreign key points at this table or out of it, so the rebuild needs
+-- neither `foreign_keys` nor `legacy_alter_table` handling. Dropping the old
+-- table takes its index with it, hence the recreate at the end.
+
+CREATE TABLE access_log_new (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    timestamp TEXT NOT NULL,
+    -- 'ticket' | 'terminal'
+    kind TEXT NOT NULL,
+    -- 'open' | 'attach' | 'detach' | 'close' | 'denied'
+    action TEXT NOT NULL,
+    -- panel account (JWT subject)
+    subject TEXT,
+    remote_ip TEXT,
+    -- system account for terminal sessions
+    ssh_user TEXT,
+    -- 'ok' | 'denied' | 'error'
+    result TEXT NOT NULL,
+    detail TEXT
+);
+
+-- `strftime` reads both shapes, so a row already written in the new one
+-- survives this unchanged apart from losing sub-second digits it does not
+-- have. A row whose timestamp is NULL — which the writer cannot produce, but
+-- the nullable column allowed — is dated now rather than at the epoch: an
+-- undatable audit record should expire a retention period from here, not be
+-- collected on the first pass after this migration.
+INSERT INTO access_log_new
+    (id, timestamp, kind, action, subject, remote_ip, ssh_user, result, detail)
+SELECT
+    id,
+    strftime('%Y-%m-%dT%H:%M:%S', COALESCE(timestamp, 'now')) || '+00:00',
+    kind, action, subject, remote_ip, ssh_user, result, detail
+FROM access_log;
+
+DROP TABLE access_log;
+ALTER TABLE access_log_new RENAME TO access_log;
+CREATE INDEX idx_access_log_timestamp ON access_log(timestamp);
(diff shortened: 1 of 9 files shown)


lollipopkit:
@coderabbitai review

Points worth attention:

- `synchronous = FULL -> NORMAL` is a durability tradeoff. Under WAL it can lose the most recent transactions to a power failure and cannot corrupt the file; the argument is that the next metrics row is seconds away. Worth a second opinion on whether that holds for the non-metrics tables sharing this database (users, config_versions, watch_tokens, ssh_known_hosts).
- `reclaim_free_pages` may run a full VACUUM on an existing database, once, to convert it to incremental. On a large file that is a blocking rewrite needing free space equal to the database size.
- The history bound now relies on lexicographic ordering of RFC 3339 text produced by the same encoder on both sides. chrono's `to_rfc3339` varies the number of fractional digits; the ordering still holds because `'+'` sorts below `'0'`, but the exact-tie case at a boundary is off by one row.

PR #1448, opened 2026-09-08
fix(diag): stop a disposed provider writing state, and scrub what a report carries

lollipopkit:
Two halves of GlitchTip issue 83. That one event was both a crash and a disclosure, so both are here.

## The crash

`servicesProvider` is keyed by the whole `Spi`. Saving a server edit makes a *different* provider and disposes this one — while the listing started before the edit is still in flight. The `state` write that comes back then throws `UnmountedRefException` out of a future nobody awaits, which reaches the zone handler and is reported as a crash. The reporter's breadcrumbs show exactly that: two `edit saved` in ten seconds, then a `service list`, then the event.

Every write below an `await` is guarded now, `finally` included — each early `return` passes through it.

`benchmark.dart` is keyed the same way and had the same defect, so it is fixed here too. Two details worth reading:

- `_store.put` stays **before** the guard. The run is going on the server either way, and that row is the only thing that lets the next open of the page find it.
- `_schedule` refuses to arm a timer once unmounted. `ref.onDispose` cancels the timer, so one armed after dispose is one nothing would ever cancel.

`pve.dart` already guards with `ref.mounted`, and `bmc.dart` bumps a generation in `onDispose`. Neither needed anything.

## The disclosure

`Spi.toString` was `Spi<user@host:port>`, and Riverpod names a family provider after its argument — so a server's address and login were quoted verbatim into the exception's text and uploaded as the issue title. Four providers are keyed by a `Spi`.

It now reads as the same token the crumbs already carry under `server`, so a report can still be followed through one machine. `displayAddr` is untouched and is what the pages call.

Under that, a `beforeSend` that takes what this install knows to be the user's back out of an outgoing event — exception values, message (formatted, template, params), breadcrumb messages and data, tags, culprit — and clears `user` and `serverName`. The reasoning is the one `SentrySink.log` already gives for dropping the log stream: a crumb is written to be published and goes through `Redact` where it is made, while an error's text is written by whoever threw it, which includes packages and Riverpod.

Still precise rather than pattern-based, for the reason `KnownIdentifiers` documents. The identifier table moves out of `CrashReport` so the report and the upload share one implementation, and gains the BMC address plus the host inside a configured URL — a dio or socket error names the host alone, not the whole address.

Const collections are replaced rather than written through: an `Unsupported operation` thrown out of `beforeSend` is caught by the SDK, which then sends the event **unscrubbed**.

## Not in this PR

The `user.ip_address` on that event is not from the SDK — `sendDefaultPii` is false, and `user` is now cleared as well. GlitchTip records it from the connection. That needs **Organization → Scrub IP Addresses** on the instance; no code change can reach it.

## Tests

- `test/services_dispose_test.dart` — answer arriving late, `ensureExec` failing late, and the ordinary path still writing. Verified against the bug: with the guards removed, the first two fail with `UnmountedRefException`.
- `test/diagnostics_scrub_test.dart` — the reported event's shape (address swapped for RFC 5737), each field the scrub covers, the const-collection trap, and that `Spi.toString` discloses nothing while still telling two machines apart.
- `test/crash_report_test.dart` follows the identifier table to its new home.

`flutter analyze` clean. Full `flutter test` green apart from `windows_install_ssh_e2e_test.dart`, which is opt-in and wants `SBM_E2E_SSH_KEY_PASSPHRASE`.

## Summary by CodeRabbit

- **Privacy**
  - Crash reports and diagnostic events now redact server names, addresses, usernames, credentials, and related metadata before submission.
  - Diagnostic details are sanitized across messages, errors, breadcrumbs, tags, transaction data, and nested fields, including map keys.
  - Server information in diagnostic output is represented by privacy-safe identifiers.
  - Events are omitted when stored identifiers cannot be loaded.

- **Bug Fixes**
  - Improved reliability when benchmark or service operations finish after leaving the screen.
  - Prevented delayed background operations from updating disposed screens or causing errors.

- **Tests**
  - Added coverage for diagnostic redaction and disposed asynchronous operations.

### DIFF
--- a/lib/core/service/crash_report.dart
+++ b/lib/core/service/crash_report.dart
@@ -6,9 +6,9 @@ import 'package:dio/dio.dart';
 import 'package:fl_lib/fl_lib.dart';
 import 'package:flutter/foundation.dart';
 import 'package:server_box/core/service/diagnostics_upload.dart';
+import 'package:server_box/core/service/known_identifiers.dart';
 import 'package:server_box/core/service/native_exit.dart';
 import 'package:server_box/core/utils/ssh_file_backend.dart';
-import 'package:server_box/data/model/server/server_private_info.dart';
 import 'package:server_box/data/res/build_data.dart';
 import 'package:server_box/data/res/store.dart';
 
@@ -93,7 +93,7 @@ abstract final class CrashReport {
     build: BuildData.build,
     os: '${Pfs.type.name} ${Platform.operatingSystemVersion}',
     locale: Platform.localeName,
-    identifiers: knownIdentifiers(Stores.server.fetch()),
+    identifiers: KnownIdentifiers.of(Stores.server.fetch()),
     previousExit: NativeExitReport.lastExit,
     previousExitTrace: NativeExitReport.lastExitTrace,
   );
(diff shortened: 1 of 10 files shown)


lollipopkit:
@coderabbitai review

Worth a close look at two things:

1. `DiagnosticsUpload.scrub` — whether any field of a `SentryEvent` that can carry user-written text is left uncovered, and whether mutating the event in place is safe for every event type `beforeSend` receives (transactions included, since `beforeSendTransaction` is unset).
2. The `ref.mounted` guards in `services.dart` and `benchmark.dart` — specifically whether any write that must happen regardless of dispose has ended up behind a guard, or the reverse.

lollipopkit:
Three findings, two applied.

**Terminal benchmark result lost after dispose — valid, fixed.** `_poll`'s blanket guard returned before `_finish`, so a poll landing as the page closed discarded the exit code and result JSON it had just fetched and left the row at `running`. The guard is per-write now: the store write happens either way, only `state` is behind it. `_schedule` and `_cleanup` already guard themselves. `test/benchmark_dispose_test.dart` covers the terminal poll, the unfinished one, and that nothing throws; restoring the blanket guard fails the first and third.

**Nested values in breadcrumb data — valid, fixed.** Substitution recurses through maps and lists now, non-strings keeping their type. It is unreachable today (`Diag.crumb` takes a `Map<String, String>`) but the SDK field is `Map<String, dynamic>`, so it was a hole waiting for a call site.

`extra`, `contexts` and `request` are still not covered, and the reasons are now in the doc comment: `extra` is deprecated in this SDK and nothing writes one; `contexts` is what `DiagnosticsPlatform` exists to decide (hardware and OS release, no name, no identifier) and holds typed objects rather than free text; `request` is set by an HTTP integration and the pure-Dart SDK has none — worth covering the day `sentry_dio` is added, not before.

**Transactions bypassing `beforeSend` — not the case here.** `SentryClient._runBeforeSend` (sentry 9.29.0, `lib/src/sentry_client.dart:545`) reads `if (event is SentryTransaction && beforeSendTransaction != null)` and falls through to `else if (beforeSend != null)` when that callback is null, which it is. A transaction therefore does reach `scrub`. Separately, nothing in this repo or in fl_lib calls `startTransaction`, and the pure-Dart SDK auto-instruments nothing, so no transaction exists to scrub. Both facts are now recorded at the registration, along with the note that spans would need their own pass if one is ever started.

Also in this push: `beforeSend` now returns null — dropping the event — when the records cannot be read at all, rather than sending it scrubbed against an empty table. And `origin/main` is merged in, cleanly.

`flutter analyze` clean. Full suite green apart from `windows_install_ssh_e2e_test.dart`, which is opt-in and wants `SBM_E2E_SSH_KEY_PASSPHRASE`.

PR #1451, opened 2026-09-09
fix: ship the az locale, allow a plain-http AI endpoint, and act on the #1450 audit

### LINKED ISSUE #1447: 希望ai功能增加http的base url的支持
Image
Image
Image目前对于本地部署模型的http协议的base URL不支持，希望增加对这个的支持

lollipopkit:
Closes #1447.

Three things in one branch: the Azerbaijani translation was shipped in a state where no user could reach it, issue #1447, and the findings from a review of #1448..#1450.

## The az translation was unreachable

`lib/generated/l10n/` is checked into this repo, and #1449 added `lib/l10n/app_az.arb` without running `flutter gen-l10n`. So there is no `l10n_az.dart`, no `Locale('az')` in `AppLocalizations.supportedLocales`, no case in `lookupAppLocalizations` — an az device resolved to English through `localeListResolutionCallback`, and Azerbaijani never appeared in the settings picker, which enumerates `supportedLocales`. The whole translation has been dead in every build since it merged.

Regenerated here. fl_lib had no `az` either, so `LibLocalizations.of` returned null for the locale and every `libL10n.*` string fell back to English — dialog buttons, `libL10n.attention`, the card titles on the benchmark result page, the unit words inside an "ago" string. That half is lollipopkit/fl_lib#52, now merged, and the gitlink moves to `fd6076d` here.

## #1447 — an AI endpoint over plain http

`https://` was always accepted and `http://localhost` always was too; everything between was refused as `invalidUrl`. That is the wrong thing to say about an address that is reachable and spelled correctly — an Ollama or vLLM box on the LAN — and it left the feature unusable for exactly the setups that have no certificate to offer.

So: a switch, not a widening. `AskAiConfig.allowInsecure`, off by default, sitting under the endpoint field in AI settings. The request carries the API key and whatever terminal output was gathered as context, and whether that network is one to send them over in the clear is not something this app can work out — same shape and same reasoning as `MonitorHttpCredential.allowInsecure`. Loopback (`localhost`, `127.0.0.1`, `[::1]`) still needs nothing.

`AskAiConfigException.insecureScheme` is separate from `invalidBaseUrl` because the two need different words: one is a typo, the other is a setting one switch away. The switch is shown whether or not the current address is `http://` — it is the answer to "why is my local model rejected", and a control that appears only after the rejected value has been saved is found after the giving up.

## Review findings

Benchmark tab:

- **The 1 s tick fired whenever a machine was selected**, which is true whenever there is a server at all, and each tick re-read `BenchmarkStore.all()` — every run of every server, logs and result documents decoded on the UI isolate, up to 50 rows a server. Now the tick compares a cheap revision (`SELECT server_id, status`) and the list is cached behind it. Both halves of that revision are needed: the count alone misses a run ending, the running set alone misses a run that starts and fails inside one second, which is what a poll answering "the run directory is gone" does. There is a test for that case.
- `_buildDetail` re-read the row by id on every rebuild, for a page that re-reads it itself. Off the cached list now.
- **Deleting a run wrote straight to the store**, so `BenchmarkState.history` kept the record and the run form went on seeding itself from a run that no longer existed. Routed through `BenchmarkNotifier.remove`, which the audit left in place with no callers at all.
- **`BenchmarkRunPage` held its `Spi` in a `late final`** under a key that a rename does not change, so the bar kept the old name for the life of the tab. Reads through `widget` now, which removes the class rather than the instance.
- The result page's `ValueKey` was redundant — its id *is* the pane's `detailId`, so `NestedNavigator` already mints a fresh page key — and its comment described a failure that cannot occur, sitting beside the one key that does matter.

Elsewhere:

- `getConnectionHistory` lost its only production caller when #1450 removed `getServerStats`. Marked `@visibleForTesting` rather than deleted: it is how a test says what is actually in the table, which the migration regression test needs — a summary cannot tell a row written wrong from one not written at all.
- `fmtDuration` moved next to `BenchmarkRun.elapsed`. Two pages were importing a list-tile widget to reach it.
- **`agent_session.g.dart`, `benchmark.g.dart` and `services.g.dart` were stale on main** — their sources were last edited in #1446 and #1448 with no codegen run. Regenerated. Only riverpod's debug hash, so nothing was broken, but the tree did not match its own generated output.
- README translation credits removed in both languages. The list is `lib/l10n/` and the people are its git history; a copy here only ever went stale — it was already missing Azerbaijani.

One finding did not hold on inspection: `GeoData.resetForTest`, `CrashLog.resetForTest` and `NativeExitReport.debugForgetPending` all already carry `@visibleForTesting`, and the codebase consistently imports it from `package:flutter/foundation.dart`, so `land.dart` is not the odd one out. The only unannotated `*ForTest` helper was fl_lib's `Update.resetForTest`, fixed in the fl_lib PR.

## Tests

Two existing ones were not testing what they said:

- `a server with no attempts is absent from the list` recorded nothing at all, so `getAllServerStats` hit its `totals.isEmpty` early return and the assertion would have passed against a stub answering `[]`.
- `firstWhere` with no `orElse` in the migration test turned a missing-rows regression into `Bad state: No element`, which names nothing and skips the three counts that would say what went wrong.

New: the pane following a change of machine (the two-column case had no coverage), reading a second past run in the pane, a run appearing between ticks, delete reaching the notifier, the run page reading its `Spi` through `widget`, and `test/ask_ai_insecure_endpoint_test.dart` — refused/allowed/loopback/https, with an adapter that refuses every request so "it reached the network" is the proof the address was accepted.

`flutter analyze` clean; `flutter test` 2542 passing. `windows_install_ssh_e2e_test.dart` fails here for want of `SBM_E2E_SSH_KEY_PASSPHRASE` and fails identically on `main` — verified by stashing.

## Summary by CodeRabbit

- **New Features**
  - Added an option to allow non-loopback HTTP endpoints for AI services, with clear security warnings.
  - Benchmark screens now update machine details more reliably and avoid unnecessary refreshes.
  - Added clearer guidance when no benchmark servers are configured.

- **Bug Fixes**
  - Improved handling of benchmark history, deletion, and rapidly completed runs.
  - Added validation for insecure, malformed, secure, and loopback AI endpoints.

- **Documentation & Localization**
  - Updated documentation to reflect support for 16 languages.
  - Added translated text for AI endpoint settings and benchmark messages.


lollipopkit:
@coderabbitai review

Worth extra attention:

- `AskAiConfig.allowInsecure` — an opt-in that permits sending an API key and terminal context over plaintext http. Check the default, the loopback exemption in `isSecureRemoteEndpoint`, and that `insecureScheme` cannot be reported for an address that is malformed rather than merely insecure.
- `_BenchmarkTabPageState`'s history cache — every path that writes `benchmark_run` has to be covered by `listRevision()` or by an explicit `_syncRevision()`, or a run silently fails to appear in the list.
- `_onDelete` reads `benchmarkProvider` only for the selected machine, deliberately: reading one into being for an unselected machine would start it polling the run being deleted.

